PR #5, opened 2016-05-09
new travis config specifying that only master and dev should be built

mkolarek:


robertmuil:
This is still building. Would be great to have the building of dependencies (scipy / numpy etc) removed from this continuous integration somehow...

mkolarek:
@robertmuil yes I am currently investigating this

mkolarek:
I have managed to reduce the build time to about 7-8min... Although this is still not ideal it at least puts the build under 10min and it doesn't time out. Possible solution is to change from pip to conda in .travis.yml.

PR #11, opened 2016-05-09
added cli module

mkolarek:


jbao:
cool, we could possibly change cli.py to expan and make setup.py install this script to some python path, such that the user only needs to type expan in the command line.

robertmuil:
+1 to Jie - and also, I don't think we need to import the cli with the core module do we? If we're importing expan within python, we don't need or want cli functionality...

mkolarek:
yes I've added it as a module without thinking too much, sorry... I can look into how setup.py can be modified, what would you guys say the priority of this is?

robertmuil:
low

On Tue, May 10, 2016 at 10:47 AM, Marko Kolarek notifications@github.com
wrote:

> yes I've added it as a module without thinking too much, sorry... I can
> look into how setup.py can be modified, what would you guys say the
> priority of this is?
> 
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> https://github.com/zalando/expan/pull/11#issuecomment-218095836

## 

_Robert Muil_ Data Scientist | +49 176 276 02775 | E-Mail:
robert.muil@zalando.de

jbao:
`setup.py` should install now the binary `expan` to some system/environment path, can someone test this?

robertmuil:
Cool! Works as advertised! That's really cool.
I'm gonna merge this to dev and then we can bump the version there before merging to master.

PR #15, opened 2016-05-15
Enhancement/version

pangeran-bottor:


robertmuil:
Hi Pangeran, thank you most kindly for contributing to the project! Wonderful to see.

I like very much your idea of reducing the number of places where the version string is hard-coded - currently we just use the `bumpversion` program to maintain these strings. I'm not 100% sure whether the regular expression parsing of the `version.py` is the easiest approach though... wouldn't it be better to just import the `version.py` module?

Also, I'll have to go through and understand where all the other commits in this pull-request are coming from - I assume that you took the `dev` branch and are now merging against `master`? (that's completely fine, but we'll have to just go through all those changes of marko and I to make sure it's all ok)...

robertmuil:
Hi Pangeran,
could you please just resubmit this PR, updating your local branch first?
I have since merged all the `dev` commits into `master`, so that if you do a PR this time for just the version change, it will be much cleaner.
Thank you very much.
Rob.

PR #16, opened 2016-05-15
Update README.rst

pangeran-bottor:
fix mock-up example to current folder names


robertmuil:
Good catch, thank you! We do need to address this namespacing stuff properly, but until we do, we'll have to use the ugly core intermediate stuff in the doc. Future work here: #18.

PR #19, opened 2016-05-16
implement expan.__version__

pangeran-bottor:
This approach is based on : http://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package/2073599#2073599


robertmuil:
awesome, thanks!

PR #22, opened 2016-05-30
only store the p-value in the chi-square test result object

jbao:


robertmuil:
This looks completely fine to me, and the tests run and coverage indicates that the changed line is covered by tests... however, confusing is that it obviously didn't fail before the fix... We probably need to fix the unittests here...

robertmuil:
In fact I think there's quite a bit problematic in the experiment unittests.
The unittest does not check the 'feature' feature at all.

This needs to be fixed... and it would also be good for us to consider that this is not detected _at all_ by the coverage tests... and I can't see how to easily fix that.

Edit:

> The feature_check currently appears to return a malformatted results structure, with percentiles unstacked as columns (instead of rows)...

scratch that - it's just the repr of the Results that unstacks percentiles.

robertmuil:
This relates to #23.

jbao:
good point, I just added a test to check for the p-values in the feature check results.

#23, however, needs some more refactoring work.

robertmuil:
Superb,  thanks Jie.
I updated the runtests script also so it's easier to just test the module under dev. (e.g. `runtests tests/tests_core/test_experiment.py`)

robertmuil:
I'll merge this as soon as CI build finishes.

Issue #2, opened 2016-05-04
Add P(uplift>0) as a statistic


jbao:
This probability will be calculated from the existing result data, based on the percentiles and the normal assumption.

robertmuil:
I am still in favour of implementing this using the result structure as it was originally intended: with percentiles in one column and value in another, using the 'uplift_pctile' statistic. However, this can be changed later if required.

Bringing over discussion from GHE:

Robert:

> two possible implementations of this:
> 1. as a completely new statistic called e.g. 'prob_greater_0', which would then also need something like 'prob_greater_cost'... and only the 'value' column would be relevant, 'pctile' would be nan.
> 2. as a pctile, with value=0 or value=. simpler, fits existing structure, and wouldn't be difficult to retrieve (res.statistic('pctile').query('value=0')

Till:

> Do not like solution 2 as it sounds very hacked (and requires previous knowledge). The query for solution 1 should also not be too difficult, as it has a name (so same as retrieving mean)

Robert:

> I don't think solution 2 is hacked: it fits exactly what the question is, namely 'what is the pctile at which 0 or cost occurs'. Putting this in as a statistic seems to me more hacked because it throws away the pctile column, and requires the statistic column to include something numeric (0 or cost) in the string.

How are we going to represent several different costs?
What if the costs arise after the calculation?
Solution 2 allows at least approx. answers to these at a later time.

jbao:
also just thought about that again, 
- encoding the probability as a percentile makes the results structure generic, but it has at least two drawbacks
  - the probabilities for different variants are persisted in the same column, which can only be reference by the variant name and a '0', not entirely impossible, but feels counter-intuitive
  - additional arithmetic operation required, since the value we store in the result is actually 1 - probability
- yes, we need to make the treatment cost as a parameter, rather than a hard-coded 0

robertmuil:
Hi Jie,
regarding the points you mentioned:
1. same column: I don't understand: each variant is a separate column...
2. additional arithmetic: yes, but this can easily be implemented in a property to expose it as though it were a statistic.
3. treatment cost: my point was rather more conceptual: what if there are several different possible treatment costs? they will all be in the result dataframe as separate statistics. this will be ugly even with a single treatment cost.
4. In addition, the treatment costs (and 0) will be encoded in a string. Which feels to me hacky.

PR #24, opened 2016-06-01
Added function to calculate P(uplift>0)

jbao:
this solves #2


expan/core/experiment.py
```diff
@@ -300,36 +300,36 @@ def delta(self, kpi_subset=None, variant_subset=None,
...
+	        relative (boolean): If relative==True, then the values will be
+	            returned as distances below and above the mean, respectively,
+	            rather than the	absolute values. In	this case, the interval is
+	            mean-ret_val[0] to mean+ret_val[1]. This is more useful in many
+	            situations because it corresponds with the sem() and std()
+	            functions.
+
+	    Returns:
+	        Results object containing the computed deltas.
+	    """
```
robertmuil, line 332: all of this whitespace change makes it hard to review a PR. Would be better in the future to make these changes in a separate commit, and merge immediately without a PR for that, I think

expan/core/results.py
```diff
@@ -4,40 +4,49 @@
...
 
 from debugging import Dbg
+from pdb import set_trace
+
+mandatory_index_levels = [
+	'metric',
+	'subgroup_metric',
+	'subgroup',
+	'statistic',
+	'pctile']
```
robertmuil: Any reason to duplicate this array?

jbao: I had to use this later on (line 334) in an external function, maybe we can make this array always global?

robertmuil: but it should already be accessible as Results.mandatory_index_levels: it's not an instance variable, just a static class variable

expan/core/experiment.py
robertmuil, line 332: any chance you could separate the whitespace from the actual change in this PR? I'm struggling to verify it... @jbao

jbao, line 332: hmm, no idea now, need to investigate.

expan/core/results.py
jbao: ah, good to know, i'll reimplement this.

expan/core/statistics.py
```diff
@@ -445,6 +445,25 @@ def normal_difference(mean1, std1, n1, mean2, std2, n2, percentiles=[2.5, 97.5],
 		return dict([(p, mean + stats.t.ppf(p / 100.0, df=d_free) * st_error)
 					 for p in percentiles])
 
+def estimate_std(x, mu, pctile, n1, n2):
+	"""Estimate the standard deviation from a given percentile, according to
+	the formula:
+		x = mu + t * sigma / sqrt(n)
```
robertmuil, line 451: that's cool. where did you get this from?

jbao, line 451: well, that's basically the inverse step of calculating the CI.

robertmuil:
I approve the merge. Haven't explored possible side-effects fully, nor verified the test cases, but it looks ok.

domheger:
I propose to add a default argument assume_normal=True to calculate_prob_uplift_over_zero() and prob_uplift_over_zero_single_metric() to make this explicit. 
If this parameter is set to False we can raise a NotImplementedError.
Well, actually we use the t-distribution here (I guess because the population std is not known)...

jbao:
@domheger that sounds like a good idea.

PR #62, opened 2017-02-01
Feature/octo 1253 expan results in json

mkolarek:
- to_json() currently works only for results created with delta() and sga(), trend() results are not supported at the moment
- current implementation is lacking, a recurrent solution would be the way to go
- unit tests have been implemented
- changes to test_experiment.py are only due to whitespace issues


expan/core/results.py
```diff
@@ -379,6 +383,65 @@ def to_hdf(self, fpath):
 
 		hfile.close()
 
+	def to_json(self):
+		"""
+
+		Args:
+		    fpath:
```
jbao: Do we want to support some filename as an argument here?

mkolarek: good idea, I can add this

jbao (approved):

PR #73, opened 2017-02-09
Feature/early stopping

jbao:
Implemented 3 early-stopping algorithms as part of the delta function:
- group sequential
- Bayes factor
- precision of the Bayes credible interval

Notes:
- `*_to_dataframe` functions need to be refactored
- more distribution models need to be added
- unit tests will now run for a while due to the addition of Bayesian methods


expan/core/early_stopping.py
```diff
@@ -0,0 +1,228 @@
...
+
+def HDI_from_MCMC(posterior_samples, credible_mass=0.95):
+    # Computes highest density interval from a sample of representative values,
+    # estimated as the shortest credible interval
+    # Takes Arguments posterior_samples (samples from posterior) and credible mass (normally .95)
+    # http://stackoverflow.com/questions/22284502/highest-posterior-density-region-and-central-credible-region
+    sorted_points = sorted(posterior_samples)
+    ciIdxInc = np.ceil(credible_mass * len(sorted_points)).astype('int')
+    nCIs = len(sorted_points) - ciIdxInc
+    ciWidth = [0]*nCIs
```
gbordyugov, line 109: switch to np.array instead of lists and get rid of the for loop below in favour of array sections?

expan/core/early_stopping.py
```diff
@@ -0,0 +1,228 @@
...
+	prior = cauchy.pdf(0, loc=0, scale=1)
+	bf = kde.evaluate(0)[0] / prior
+	stop = int(bf > 3 or bf < 1/3.)
+
+	interval = HDI_from_MCMC(traces['alpha'])
+
+	return stop, mu_x-mu_y, {'lower':interval[0],'upper':interval[1]}, n_x, n_y, mu_x, mu_y
+
+
+def bayes_precision(x, y, distribution='normal', posterior_width=0.08):
```
gbordyugov, line 198: looks like duplication in comparison to above

expan/core/experiment.py
```diff
@@ -467,6 +517,144 @@ def delta(self, kpi_subset=None, derived_kpis=None, variant_subset=None,
 
 		return res
 
+	def group_sequential_delta(self,
```
gbordyugov, line 518: `group_sequential_delta`, `bayes_factor_delta`, and `bayes_precision_delta` bear a lot of duplicate code - any hope of factoring out a common pattern?

expan/core/early_stopping.py
```diff
@@ -0,0 +1,228 @@
...
+					'x': _x, 
+					'y': _y}
+	else:
+		raise NotImplementedError
+	model_file = __location__ + '/../models/' + distribution + '_kpi.stan'
+	sm = StanModel(file=model_file)
+
+	fit = sm.sampling(data=fit_data, iter=25000, chains=4, n_jobs=1, seed=1)
+	traces = fit.extract()
+	interval = HDI_from_MCMC(traces['delta'])
```
gbordyugov: maybe:

`interval[0], interval[1] = min(interval[0], interval[1]), max(interval[0], interval[1])`
to ensure the sanity of interval bounds?

expan/core/early_stopping.py
```diff
@@ -0,0 +1,228 @@
...
+					'x': _x, 
+					'y': _y}
+	else:
+		raise NotImplementedError
+	model_file = __location__ + '/../models/' + distribution + '_kpi.stan'
+	sm = StanModel(file=model_file)
+
+	fit = sm.sampling(data=fit_data, iter=25000, chains=4, n_jobs=1, seed=1)
+	traces = fit.extract()
+	kde = gaussian_kde(traces['delta'])
```
gbordyugov: the difference between this and `bayes_precision` is only the following 6 lines of code - please correct me if I'm wrong - we could think about factoring out the common pieces of the code between those two funcs

expan/core/experiment.py
```diff
@@ -360,9 +361,58 @@ def __str__(self):
...
+			return self.fixed_horizon_delta(kpi_subset, derived_kpis, **kwargs)
+		elif method == 'group_sequential':
+			return self.group_sequential_delta(res, kpis_to_analyse, **kwargs)
+		elif method == 'bayes_factor':
+			return self.bayes_factor_delta(res, kpis_to_analyse, **kwargs)
+		elif method == 'bayes_precision':
+			return self.bayes_precision_delta(res, kpis_to_analyse, **kwargs)
+		else:
+			raise NotImplementedError
+
```
gbordyugov:
```
method_table = {
  'fixed_horizon': (self.fixed_horizon_delta, kpi_subset, derived_kpis),
  'group_sequential': (self.group_sequential_delta, res, kpis_to_analyse),
  'bayes_factor': (self.bayes_factor_delta, res, kpis_to_analyse),
  'bayes_precision': (self.bayes_precision_delta, res, kpis_to_analyse)
}

if not method in method_table:
  raise NotImplementedError
else:
  m = method_table[method]
  f, a1, a2 = m[0], m[1], m[2]
  f(a1, a2, **kwargs)
```

would be easier to expand in feature

expan/core/early_stopping.py
```diff
@@ -0,0 +1,228 @@
...
+					 information_fraction=1, 
+					 alpha=0.05, 
+					 cap=8):
+	"""
+	Group sequential method to determine whether to stop early or not.
+
+	Args:
+		x (array_like): sample of a treatment group
+		y (array_like): sample of a control group
+		spending_function: name of the alpha spending function, currently
```
gbordyugov, line 39: that can be the spending function itself instead of the string with its name, see `callable(spending_function)` call below

expan/core/early_stopping.py
```diff
@@ -0,0 +1,228 @@
...
+	_y = np.array(y, dtype=float) 
+
+	# if scalar, assume equal spacing between the intervals
+	#if not isinstance(information_fraction, list): 
+	#	fraction = np.linspace(0,1,information_fraction+1)[1:]
+	#else:
+	#	fraction = information_fraction
+
+	# alpha spending function
+	if spending_function in ('obrien_fleming'):
```
gbordyugov, line 71: continuing the comment above about passing the function itself instead of its name:

```
if callable(spending_function):
  alpha_new = spending_function(information_fraction, alpha = alpha)
else:
  raise NotImplemented
```

`callable()` works only in Python 2.x and Python 3.2+ though

expan/core/results.py
```diff
@@ -678,6 +678,58 @@ def feature_check_to_dataframe(metric,
...
+	if len(pctiles) > 0:
+		df = df.append(pd.DataFrame({
+			'metric': metric,
+			'statistic': 'uplift_pctile',
+			'pctile': list(pctiles.keys()),
+			'value': list(pctiles.values()),
+			'subgroup_metric': subgroup_metric,
+			'subgroup': subgroup
+			}))
+
```
gbordyugov, line 738: Can this duplication be avoided by the Dict.update() method - am I overlooking smth?

gbordyugov (commented):
great job Jie!

my comments are mostly stylistic grumblings

expan/core/results.py
```diff
@@ -678,6 +678,58 @@ def feature_check_to_dataframe(metric,
...
+	    subgroup_metric: subgroup metric name
+	    subgroup: subgroup name
+
+	Returns:
+		the result data frame
+	"""
+	df = pd.DataFrame({
+		'metric': metric,
+		'pctile': None,
+		'statistic': ['uplift', 'sample_size', 'variant_mean', 'stop'],
```
gbordyugov, line 724: Don't you think that we might lift `stop` to the level of `metric`? This
a) will be more compliant to the doc-string above (line 697)
b) will make exporting data frame to other formats smoother

gbordyugov (commented):
location of `stop` label in the hierarchy of data frame indices

PR #80, opened 2017-02-23
Early stopping

jbao:
- Add a Poisson KPI model for Stan and the unit test (**Unit tests might run even longer now**)
- Fixed `setup.py` such that Stan model files are also installed with the remaining modules


tests/tests_core/test_early_stopping.py
```diff
@@ -70,13 +72,27 @@ def test_bayes_factor(self):
 		stop,delta,CI,n_x,n_y,mu_x,mu_y = es.bayes_factor(self.rand_s1, self.rand_s2)
 		self.assertEqual(stop, 1)
 		self.assertAlmostEqual(delta, -0.15887364780635896)
-		self.assertAlmostEqual(CI['lower'], -0.24414725578976518)
-		self.assertAlmostEqual(CI['upper'], -0.072120687308212819)
+		self.assertAlmostEqual(CI['lower'], -0.25165623415486293)
```
mkolarek: this assert seems to fail
E           AssertionError: -0.24873596733759926 != -0.25165623415486293 within 7 places

mkolarek: and the one at line number 109
E           AssertionError: -0.24873596733759926 != -0.25165623415486293 within 7 places

jbao:
ok, obviously the addition of a prior messed up with the random seed, fixed that now. Does that resolve #85 , @gbordyugov ?

gbordyugov:
unfortunately not @jbao

jbao:
Reducing the initial `stepsize` seems to solve the problem of numerical instability (which amounts to several run-time warnings about parameter constraints).

See also https://groups.google.com/forum/#!topic/stan-users/hpGvb_ZpJ3c

Issue #81, opened 2017-02-27
Bad code duplication in experiment.py


gbordyugov:
`Experiment.delta()` is now a dispatcher, however those lines https://github.com/zalando/expan/blob/dev/expan/core/experiment.py#L368 duplicate the code https://github.com/zalando/expan/blob/dev/expan/core/experiment.py#L455

Additional complication: `delta()` function calls the other one, i.e. `fixed_horizon_delta()`

chaitanyasd:
I'm a newbie. Can i take this? Please can you help.

jbao:
@csdeshpande19 sure, this requires some de-duplication of the `delta()` and `fixed_horizon_delta()` function. Let us know if you have any specific questions.

chaitanyasd:
Can you tell me what exactly I have to fix so i can start working on it.

jbao:
My idea would be to remove the duplicated code in `fixed_horizon_delta()` and do all the pre-processing in `delta()`,  but feel free to open a PR and maybe it would become easier to discuss. What do you think?

chaitanyasd:
Yeah. No problem. I'll do a PR in few hours as I'm new to Open Source so it may take some time to get acquainted with it. I'll try to help as much as I can.

s4826:
Does this still need to be done?

jbao:
Hi @s4826 , yes, this one is still open AFAIK.

s4826:
Ok, I'll open a pull request.

PR #86, opened 2017-03-13
Test

s4826:
- Remove duplicated code in ```fixed_horizon_delta()```
- updated tests to reflect new method signature for ```fixed_horizon_delta()```

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -385,6 +385,7 @@ class Experiment(ExperimentData):
 
 		if kpi_subset is not None:
 			kpis_to_analyse.intersection_update(kpi_subset)
+		res.metadata['kpis_to_analyse'] = kpis_to_analyse
 		self.dbg(3, 'kpis_to_analyse: ' + ','.join(kpis_to_analyse))
 
 		method_table = {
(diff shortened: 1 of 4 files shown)


jbao (approved):

mkolarek:
:+1:

mkolarek (approved):

expan/core/experiment.py
```diff
@@ -399,9 +400,14 @@ def delta(self, method='fixed_horizon', kpi_subset=None, derived_kpis=None, **kw
...
 			f, a1, a2 = m[0], m[1], m[2]
-			return f(a1, a2, **kwargs)
+			if method == 'fixed_horizon':
+				# pass initialized results object into fixed_horizon_delta
+				return f(res, a1, a2, **kwargs)
+			else:
+				return f(a1, a2, **kwargs)
 
-	def fixed_horizon_delta(self, 
+	def fixed_horizon_delta(self,
```
gbordyugov, line 409: Is introducing an additional parameter to fixed_horizon_delta() worth it? It seems to me that this introduces an inhomogeneity across the fourth `delta` workers

tests/tests_core/test_results.py
```diff
@@ -37,6 +38,34 @@ def load_example_results():
 
 	return r.from_hdf(example_fpath)
 
+def mock_results_object(data, kpi_subset=None, derived_kpis=None, **kwargs):
+	"""
+	Create a results object for any tests involving fixed_horizon_delta()
```
gbordyugov, line 43: why factor out this functionality as a separate function? is it going to be used anywhere besides `fixed_horizon_delta()`?

addendum: would this code persist in Experiment.delta() further?

gbordyugov (changes requested):
Hi,

great job! I'm however still a bit unconvinced about your solution to move the duplicate code into the `mock` function. Does it eliminate the duplicate code in both `delta()` and `fixed_horizon_delta()`?

Plus introducing a `res` argument to `fixed_horizon_delta()` makes its handling a bit special in comparison to other workers.

Best,
gb

gbordyugov (approved):
Hi,

I totally misunderstood the logic of what you did, now, after a bit of pondering, it looks great to me.

jbao:
This resolves #81 .

PR #104, opened 2017-04-19
OCTO-1502 removed variant_subset parameter...

gbordyugov:
OCTO-1502

... from the args list of `Experiment.fixed_horizon_delta()`

In the current implementation of `Experiment.fixed_horizon_delta()`, the parameter `variant_subset` and all derivatives of it are ignored - hence, it's safe to remove it from the argument list.


shansfolder:
👍

mkolarek (approved):

PR #110, opened 2017-04-24
OCTO-1502: cleanup of call chains

gbordyugov:
OCTO-1502

introducing `make_delta()` to hide statistics parameters in the call chains.


gbordyugov:
hmm, it fails with minor diffs in floating point constants in low-level bayesian functions that I didn't touch...

mkolarek:
I would say that the difference is negligible, from my point of view you can simply update the unit test values.

mkolarek (approved):

gbordyugov:
I'm running `tox`-based tests locally in order to reproduce (just used `runtests` before)

expan/core/experiment.py
```diff
@@ -383,33 +330,36 @@ def delta(self, method='fixed_horizon', kpi_subset=None, derived_kpis=None, **kw
...
 				# TODO: only works for ratios
 				res.metadata['reference_kpi'][kpiName] = re.sub(pattern+'/', '', dk['formula'])
+				reference_kpis[kpiName] = re.sub(pattern+'/', '', dk['formula'])
 
 		if kpi_subset is not None:
 			kpis_to_analyse.intersection_update(kpi_subset)
 		self.dbg(3, 'kpis_to_analyse: ' + ','.join(kpis_to_analyse))
 
+		defaultArgs = [res, kpis_to_analyse]
+		deltaWorker = statx.make_delta(assume_normal, percentiles, min_observations,
```
shansfolder, line 340: This ```deltaWorker``` seems not used in the current method.

shansfolder: Should we pass ```deltaWorker``` to ```fixed_horizon_delta```? Maybe that's the reason why unit test is failed.

expan/core/experiment.py
```diff
@@ -383,33 +330,36 @@ def delta(self, method='fixed_horizon', kpi_subset=None, derived_kpis=None, **kw
...
+				                       nruns, relative)
 		method_table = {
-			'fixed_horizon':    self.fixed_horizon_delta,
-			'group_sequential': self.group_sequential_delta,
-			'bayes_factor':     self.bayes_factor_delta,
-			'bayes_precision':  self.bayes_precision_delta,
+			'fixed_horizon':    (self.fixed_horizon_delta,    defaultArgs + [reference_kpis, weighted_kpis]),
+			'group_sequential': (self.group_sequential_delta, defaultArgs                                  ),
```
shansfolder: I am wondering should ```group_sequential_delta``` also take a deltaWorker parameter?

expan/core/statistics.py
```diff
@@ -9,6 +9,13 @@ def _delta_mean(x, y):
 	"Implemented as function to allow calling from bootstrap."
 	return np.nanmean(x) - np.nanmean(y)
 
+def make_delta(assume_normal=True, percentiles=[2.5, 97.5],
```
shansfolder, line 12: I am a little not satisfied with the name ```make_delta```, because it might cause confusion with ```delta```, which also makes a delta with additional parameters. 
but couldn't think of a better name at the moment too.

shansfolder (changes requested):

expan/core/experiment.py
gbordyugov: you're correct, I overlooked that

expan/core/experiment.py
gbordyugov: nopes, it has an early stopping low-level worker

shansfolder:
Thank you Grisha. It looks good! See my comments above for a few small suggestions.

expan/core/experiment.py
gbordyugov: The thing is the following: most old `delta()`-like methods (`fixed_horizon_delta()`, `sga()`, `trend()`, `feature_check()`, etc., use the same low-level `delta()` backend and thus share its common numerical parameters. That's why it seems to be a good idea to pass a partially applied `delta()` around (in the form of `deltaWorker`) and thus spare the number of arguments.

All new Bayesian funcs have a separate low-level statistical worker. In that sense, the umbrella func `delta()` relies on the default parameters of those and passing a partially applied worker would not make much sense.

We have to think about decoupling the old `*_delta()` code from the new one.

expan/core/statistics.py
gbordyugov: If `delta()` is a function, then `make_delta()` creates a `delta()` function on the fly and returns it -- seems logical to me :~)

shansfolder: oh yes. you are right!

shansfolder (approved):

PR #115, opened 2017-05-24
OCTO-1614 cleanup module structure

shansfolder:
What I did in this PR:
* Move some root-level functions to class methods
* Move other root-level functions to util module
* Optimize imports and cleanup code
* Restructure test folders


shansfolder:
Have no idea why unit test failed on Python 2.7. Need to investigate further.

shansfolder:
Ok I fixed the build. 

The reason is that pytest just released a new version **3.1.0** on May. 22th(see changelog [here](https://docs.pytest.org/en/latest/changelog.html#new-features)), which leads to our failed tests with Python 2.7.
As a hot fix, I fix pytest version to **3.0.7** when running tox.

gbordyugov:
looks ok to me, especially considering the `+5,677 −5,874` thing ;-)

PR #126, opened 2017-06-20
OCTO-1804: Optimize the loading of .stan model in expan.

daryadedik:
- Stan models are loaded/compiled only once and saved as .pkl files in the models folder and reused every time they are needed.
- For unit tests the .pkl files are deleted afterwards with the tearDownModule() function.


tests/tests_core/test_experiment.py
```diff
@@ -5,10 +5,15 @@
 
 from expan.core.experiment import Experiment
 from expan.core.results import Results
-from expan.core.util import generate_random_data
+from expan.core.util import generate_random_data, remove_model_pkls
 from tests.tests_core.test_results import mock_results_object
 
 
+# Removes .pkl compiled model files after the module tests are finished
+def tearDownModule():
```
shansfolder: It would be great if you use triple quotes (docstring) for documenting method...

shansfolder: and it would be good to add a sentence why tearDownModule is inside one class.

expan/core/util.py
```diff
@@ -151,3 +154,12 @@ def generate_random_data_n_variants(n_variants=3):
     }
 
     return test_data_frame, metadata
+
+
+# Removes .pkl compiled model files
+def remove_model_pkls():
```
shansfolder: Use docstring here too.

shansfolder (changes requested):
Build time of only 6 mins! Great work ;)

tests/tests_core/test_experiment.py
daryadedik: Good points!

shansfolder (approved):

PR #131, opened 2017-06-30
Stan models compilation, exceptions catch, unit tests adaptation.

daryadedik:
1. Re-implemented Stan model load procedure: .pkl files (Stan models are pickable) with the compiled models are saved in the temporary folder defined by tempfile.gettempdir() which depends on the platform/os and settings.
2. Added check/exception if kpi_subset is not a list of strings.
3. Changed bayes-specific unit tests (to_json() tests) in order to pass fewer number of iterations.

### DIFF
--- a/expan/core/early_stopping.py
+++ b/expan/core/early_stopping.py
@@ -9,6 +9,7 @@ from expan.core.util import drop_nan
 import expan.core.statistics as statx
 
 import pickle
+import tempfile
 
 __location__ = realpath(join(os.getcwd(), dirname(__file__)))
 
@@ -125,6 +126,37 @@ def HDI_from_MCMC(posterior_samples, credible_mass=0.95):
     return (HDImin, HDImax)
 
 
+def compile_stan_model(model_file, distribution):
+    """
+    Creates Stan model. Compiles a Stan model and saves it to .pkl file to the /tmp folder if
+        file doesn't exist yet and load precompiled model if there is a model file in the /tmp.
+    Args:
+        model_file: model file location
+        distribution: name of the KPI distribution model, which assumes a 
+            Stan model file with the same name exists
+    Returns:
+        returns compiled Stan model for the selected distribution or normal distribution
+            as a default option
+    Note: compiled_model_file is the hardcoded file path which may cause some issues in future.
+    There are 2 alternative implementations for Stan models handling:
+        1. Using global variables
+        2. Pre-compiling stan models and adding them as a part of expan project
+        (3). Using temporary files with tempfile module is not currently possible, since it 
+            generates a unique file name which is difficult to track.
+        However, compiled modules are saved in temporary directory using tempfile module 
+        which vary based on the current platform and settings. Cleaning up a temp dir is done on boot.
+    """
+    compiled_model_file = tempfile.gettempdir() + '/expan_early_stop_compiled_stan_model_' + distribution + '.pkl'
+
+    if os.path.isfile(compiled_model_file):
+        sm = pickle.load(open(compiled_model_file, 'rb'))
+    else:
+        sm = StanModel(file=model_file)
+        with open(compiled_model_file, 'wb') as f:
+            pickle.dump(sm, f)
+    return sm
+
+
 def _bayes_sampling(x, y, distribution='normal', num_iters=25000):
     """
     Helper function.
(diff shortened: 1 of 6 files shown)


expan/core/early_stopping.py
```diff
@@ -125,6 +126,37 @@ def HDI_from_MCMC(posterior_samples, credible_mass=0.95):
...
+        1. Using global variables
+        2. Pre-compiling stan models and adding them as a part of expan project
+        (3). Using temporary files with tempfile module is not currently possible, since it 
+            generates a unique file name which is difficult to track.
+        However, compiled modules are saved in temporary directory using tempfile module 
+        which vary based on the current platform and settings. Cleaning up a temp dir is done on boot.
+    """
+    compiled_model_file = tempfile.gettempdir() + '/expan_early_stop_compiled_stan_model_' + distribution + '.pkl'
+
+    if os.path.isfile(compiled_model_file):
```
gbordyugov, line 151: Is there a way to ensure that `compiled_model_file` is the same file each time you call this function?

expan/core/experiment.py
```diff
@@ -76,6 +75,10 @@ def delta(self, method='fixed_horizon', kpi_subset=None, derived_kpis=None,
         'bayes_factor':			self.bayes_factor_delta()
         'bayes_precision':		self.bayes_precision_delta()
         """
+        # Check the correct type of the kpi_subset which should be a list
+        if not isinstance(kpi_subset, list) and not all(isinstance(kpi, str) for kpi in kpi_subset):
+            raise TypeError('KPI subset should be a list of strings')
+
```
gbordyugov, line 81: why do we need it here (sorry for stupid questions ;-)

gbordyugov (changes requested):

expan/core/experiment.py
daryadedik: There are no stupid questions =) This is a check whether kpi_subset should be a list of strings because a user can provide a string of a kpi to delta if there is only 1 kpi he want to use, but he should provide a list with 1 kpi, otherwise test fails. This was also added on backend request.

expan/core/early_stopping.py
daryadedik: according to documentation it should be the same because it depends on the platform and does not change within limits of one machine (but there is nowhere directly said that in documentation). I checked the path from different python consoles and it is the same.

gbordyugov (approved):

expan/core/early_stopping.py
```diff
@@ -125,6 +126,37 @@ def HDI_from_MCMC(posterior_samples, credible_mass=0.95):
     return (HDImin, HDImax)
 
 
+def compile_stan_model(model_file, distribution):
```
shansfolder: Is ```get_or_compile_stan_model``` a better name for the function?

expan/core/experiment.py
```diff
@@ -384,38 +380,33 @@ def bayes_precision_delta(self,
...
-                df.loc[(mname, '-', slice(None), 'stop'), ('value', self.baseline_variant)] = 0
-
-                if result.df is None:
-                    result.df = df
-                else:
-                    result.df = result.df.append(df)
-        finally:
-            # Remove .pkl compiled stan model files in the finally no matter script in try was successful or not
-            if remove_pkls:
-                remove_model_pkls()
```
shansfolder, line 418: Just to comfirm: so we don't remove pickled file in tmp?

expan/core/early_stopping.py
daryadedik: create_stan_model maybe would be shorter?

shansfolder: It's not about size. This method does two things, load the precompiled model if existed or create the model. 
I would prefer to name it as ```action1_or_action2``` to be clearer what it does. What do you think?

Not sure about being "Pythonic", but this is a common practice in Java community.
e.g. 
[getOrDefault](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableMap.html#getOrDefault-java.lang.Object-V-) in Google Guava
[getOrCreate](https://spark.apache.org/docs/2.0.1/api/java/org/apache/spark/sql/SparkSession.Builder.html#getOrCreate()) in Apache Spark.

expan/core/experiment.py
daryadedik: as far as I am aware, it's not a good practice to remove temporary files from /tmp or /var/folders. They are usually cleaned up after a reboot (this procedure depends on the OS). If we need to delete them manually we should think whether its ok/safe to do that or decide about using some another procedure.

expan/core/early_stopping.py
daryadedik: I agree! Will change that. Thank you!

expan/core/experiment.py
shansfolder: ok got it ;)

shansfolder (approved):

expan/core/experiment.py
daryadedik: I just investigated this question, and now I think it should be safe to delete files from /var/folders/ because you're aware of files you created. Whether to delete compiled models or not depends on how often they will/should be changed or will they be changed at all in expan.

shansfolder: they won't change for the same installation of ExpAn, but can be different once the user is using a new version of ExpAn next time.

mkolarek (approved):

PR #134, opened 2017-07-17
Octo 1616 no experimentdata

gbordyugov:
* PLEASE DON'T MERGE YET *


expan/core/__init__.py
```diff
@@ -3,7 +3,8 @@
 
 from __future__ import absolute_import
 
-__all__ = ["binning", "experiment", "experimentdata", "results", "statistics", "util", "version"]
+# __all__ = ["binning", "experiment", "experimentdata", "results", "statistics", "util", "version"]
+__all__ = ["binning", "experiment", "statistics", "util", "version"]
```
shansfolder, line 7: 💯

expan/core/early_stopping.py
```diff
@@ -159,6 +165,8 @@ def get_or_compile_stan_model(model_file, distribution):
             pickle.dump(sm, f)
     return sm
 
+cacheSamplingResults = False
+samplingResults = {} # memoized sampling results
```
shansfolder: Just occur to my mind: we might need a guideline of naming convention. e.g. when to use ```variableName``` or ```variable_name```.

expan/core/early_stopping.py
```diff
@@ -228,26 +250,37 @@ def bayes_factor(x, y, distribution='normal', num_iters=25000):
...
+    credibleMass = 0.95                # another magic number
+    leftOut      = 1.0 - credibleMass
+    p1           = round(leftOut/2.0, 5)
+    p2           = round(1.0 - leftOut/2.0, 5)
+    interval = HDI_from_MCMC(traces['delta'], credibleMass)
+
+    # return stop, mu_x - mu_y, {'lower': interval[0], 'upper': interval[1]}, n_x, n_y, mu_x, mu_y
+    return {'stop'     : stop,
+            'delta'    : mu_x - mu_y,
+            'interval' : {p1*100: interval[0], p2*100: interval[1]},
```
shansfolder: A small thing: should we convert key ```p1*100``` to string with two digits?

expan/core/experiment.py
```diff
@@ -20,396 +24,103 @@
...
-        # Add baseline to metadata
-        self.metadata['baseline_variant'] = baseline_variant
-
-
-    @property
-    def baseline_variant(self):
-        """
-        Returns the baseline variant.
+    def __init__(self, controlVariantName, data, metadata, reportKpiNames=None, derivedKpis=[]):
+        reportKpiNames = reportKpiNames or getColumnNamesByType(data, np.float64)
```
shansfolder: so if no ```reportKpiNames``` passed, we use all numerical columns as kpis?

shansfolder: placeholder (maybe I will find later when continue reading): where do we set up ```features``` now?

expan/core/experiment.py
```diff
@@ -20,396 +24,103 @@
...
-        return self.metadata['baseline_variant']
+        derivedKpiNames    = [k['name']    for k in derivedKpis]
+        derivedKpiFormulas = [k['formula'] for k in derivedKpis]
 
+        experimentColumnNames = ['entity', 'variant']
 
-    def __str__(self):
-        res = super(Experiment, self).__str__()
+        # what columns do we expect to find in the data frame?
+        requiredColumnNames = (set(reportKpiNames) | set(experimentColumnNames)) - set(derivedKpiNames)
```
shansfolder: ```|``` means union of sets here right? That's pretty cool.

expan/core/experiment.py
```diff
@@ -20,396 +24,103 @@
...
+                treatmentWeight = self._getWeights(kpi, variant)
+                ds = worker(x=treatment*treatmentWeight, y=control*controlWeight)
+                result[kpi][variant] = {'controlVariant'   : self.controlVariantName,
+                                        'treatmentVariant' : variant,
+                                        'deltaStatistics'  : ds}
 
         return result
 
+##
+##            !!!! HERE BE DRAGONS !!!!
```
shansfolder: ??

expan/core/experiment.py
```diff
@@ -20,396 +24,103 @@
...
+            result[kpi] = {}
+            control       = self.getKPIbyNameAndVariant(kpi, self.controlVariantName)
+            controlWeight = self._getWeights(kpi, self.controlVariantName)
+            for variant in self.variantNames:
+                treatment       = self.getKPIbyNameAndVariant(kpi, variant)
+                treatmentWeight = self._getWeights(kpi, variant)
+                ds = worker(x=treatment*treatmentWeight, y=control*controlWeight)
+                result[kpi][variant] = {'controlVariant'   : self.controlVariantName,
+                                        'treatmentVariant' : variant,
+                                        'deltaStatistics'  : ds}
```
shansfolder: do we still take into account single statistics (e.g. variant sample size, data mean, etc) in the result object?

shansfolder (commented):
Thank you for the great job. Code looks a lot clearer!

expan/core/early_stopping.py
gbordyugov: we could do that, but floats look more compact to me and I don't see that they don't suffice

expan/core/experiment.py
gbordyugov: ++

expan/core/experiment.py
gbordyugov: yes, they're part of `'deltaStatistics' : ds`

expan/core/experiment.py
gbordyugov: everything below that is remnants of the old code and is broken

expan/core/early_stopping.py
gbordyugov: I'm not quite sure here. My solution is just the first thing that occurred to me. Not doing rounding results in numbers like `97.50000000002` which are pain in the back to type in later

expan/core/experiment.py
gbordyugov: 1) yes, that was my first solution, maybe include `int` columns too, we can discuss this

2) there are no features yet. I'm not sure that I'm convinced to store features in a different manner than KPIs.

expan/core/experiment.py
daryadedik: just a clarification: what is a data structure of ds? data frame, array?

gbordyugov: Looking at the code, `worker` is a closure, containing either:
 - `statistics.delta()`, or
 - `early_stopping.group_sequential()`, or
 - `early_stopping.bayes_factor()`, or
 - `early_stopping.bayes_precision()`.

Looking at the definitions of those, you'll see that they basically return just a dictionary.

shansfolder:
Is this PR ready for the final review? :)

expan/core/experiment.py
daryadedik: aha, apologies, thought it was tuple as before. You change it to dictionary, good! Thanks!

expan/core/early_stopping.py
daryadedik: According to PEP8:

Function Names
Function names should be lowercase, with words separated by underscores as necessary to improve readability.
mixedCase is allowed only in contexts where that's already the prevailing style, to retain backwards compatibility.

Method Names and Instance Variables
Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability.
Use one leading underscore only for non-public methods and instance variables.

Class Names
Class names should normally use the CapWords convention.

We may need to look which style is prevailing to not change much.

shansfolder: yeah that sounds good.
I am fine with all, we just need to make a guideline and stick to it, so that the code will become more consistent and thus more readable.

daryadedik: ...and argument names also should be lowercased for PEP8 =)

expan/core/experiment.py
```diff
@@ -64,750 +84,40 @@ def __str__(self):
...
-
-        method_table = {
-            'fixed_horizon': (self.fixed_horizon_delta, defaultArgs + [deltaWorker]),
-            'group_sequential': (self.group_sequential_delta, defaultArgs + [alpha]),
-            'bayes_factor': (self.bayes_factor_delta, defaultArgs),
-            'bayes_precision': (self.bayes_precision_delta, defaultArgs),
-        }
-
-        if not method in method_table:
+        if not method in workerTable:
```
daryadedik: `if method not in workerTable` may be a bit more readable.

tests/tests_core/test_early_stopping.py
```diff
@@ -57,16 +57,16 @@ def test_group_sequential(self):
...
+        res = es.group_sequential(self.rand_s1, self.rand_s2)
+
+        self.assertEqual               (res['stop'],            True)
+        self.assertAlmostEqual         (res['delta'],          -0.15887364780635896)
+        np.testing.assert_almost_equal (res['interval'][02.5], -0.24461812530841959, decimal=5)
+        np.testing.assert_almost_equal (res['interval'][97.5], -0.07312917030429833, decimal=5)
+        self.assertEqual               (res['n_x'],             1000)
+        self.assertEqual               (res['n_y'],             1000)
+        self.assertAlmostEqual         (res['mu_x'],           -0.045256707490195384)
+        self.assertAlmostEqual         (res['mu_y'],            0.11361694031616358)
```
daryadedik, line 69: to my opinion this structure is rather difficult to maintain if there are some changes (you need to keep in mind spaces all the time).
https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

mkolarek: I agree, it would be easier if we could agree on a specific style.

expan/core/early_stopping.py
mkolarek: +1 for PEP8

mkolarek (approved):
looks great, we could potentially merge these changes with the dev branch, but not with master. I believe this would ease further collaboration without breaking and outside dependencies (yet).

expan/core/experiment.py
```diff
@@ -20,396 +24,103 @@
...
+        for formula in derivedKpiFormulas:
+            names = re.findall(kpiNamePattern, formula)
+            requiredColumnNames = requiredColumnNames | set(names)
 
-        variants = self.variant_names
+        for c in requiredColumnNames:
+            if c not in data:
+                raise ValueError('No column %s provided'%c)
 
-        res += '\n {:d} variants: {}'.format(len(variants),
```
daryadedik: "res" is unresolved variable, there are no declaration (or was commented?)

gbordyugov: I think it's part of the old (removed) code, hence in red.

daryadedik: got it! thanks!

expan/core/experiment.py
```diff
@@ -20,396 +24,103 @@
...
-                    if res.df is None:
-                        res.df = df
-                    else:
-                        res.df = res.df.append(df)
-
-            except ValueError as e:
-                res.metadata['errors']['Experiment.delta'] = e
+        variants = self.variant_names
 
+        res += '\n {:d} variants: {}'.format(len(variants),
```
daryadedik: Ah, here the res was commented and now 'res' has no declaration. Comment before actually was about that, I just missed the place.

gbordyugov:
variable names style change from camelCase to snake_case

tests/tests_core/test_early_stopping.py
gbordyugov: I agree that writing this might be a bit cumbersome, but reading is much easier - and we tend to read more often than write.

PR #140, opened 2017-08-01
Small improvement on default type of report_kpi_names

shansfolder:
To make it explicit that input argument should be a list because

```report_kpi_names='kpi'``` will lead to ```set(report_kpi_names)==set('k', 'p', 'i)```.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -37,6 +37,9 @@ class Experiment(object):
         if type(report_kpi_names) is str:
             report_kpi_names = [report_kpi_names]
 
+        if type(report_kpi_names) is not list:
+            raise TypeError('report_kpi_names should be a list of str')
+
         if report_kpi_names:
             report_kpi_names_needed = set(report_kpi_names)
         else:
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -30,20 +30,20 @@ class Experiment(object):
     """
     Class which adds the analysis functions to experimental data.
     """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by gbordyugov** [comment]
Great, this would certainly give an explicit hint to pass a list instead of smth (but still wouldn't prevent passing a single string in place of a list)

**Comment by gbordyugov** [comment]
I think it's a great addition, making the signature of the constructor more explicit.

Would be nice to have a unit test or two to be able catching bugs here.

File: expan/core/experiment.py [hunk 1]
```diff
@@ -30,20 +30,23 @@ class Experiment(object):
     """
     Class which adds the analysis functions to experimental data.
     """
-    def __init__(self, control_variant_name, data, metadata, report_kpi_names=None, derived_kpis=[]):
+    def __init__(self, control_variant_name, data, metadata, report_kpi_names=[], derived_kpis=[]):
         experiment_column_names = set(['entity', 'variant'])
         numerical_column_names  = set(get_column_names_by_type(data, np.number))
 
+        if type(report_kpi_names) is str:
+            report_kpi_names = [report_kpi_names]
```
gbordyugov, line 38: yes, totally makes sense

maybe a unit test case for that?

gbordyugov (changes requested):

expan/core/experiment.py
shansfolder: done

gbordyugov (approved):
look great!

PR #144, opened 2017-08-02
Fix delta/alpha model para inconsistency

shansfolder:
In Stan model file, I use ```delta``` to represent absolute effect size, and ```alpha``` for the normalized one, (swapped between the two), so that it is now consistent with the notation in regular delta.

In bayes factor, use trace of ```alpha``` to determine the threshold, but use trace of ```delta``` for credible interval in result (to be the same as results in regular delta).


daryadedik (approved):

expan/core/early_stopping.py
```diff
@@ -260,8 +260,10 @@ def bayes_factor(x, y, distribution='normal', num_iters=25000):
         dictionary with statistics
     """
     traces, n_x, n_y, mu_x, mu_y = _bayes_sampling(x, y, distribution=distribution, num_iters=num_iters)
-    kde = gaussian_kde(traces['delta'])
+    trace_to_evaluate_bf = traces['alpha'] if distribution == 'normal' else traces['delta']
+    trace_to_compute_credible_interval = traces['delta']
```
gbordyugov: I don't quite understand this diff between being normal and not and why we use different traces depending on that

shansfolder: The problem is in the models, ```normal.stan``` has ```alpha``` and ```delta```, but cauchy.stan only has ```delta```. I renamed the paras such that ```alpha``` is always the normalized ```delta```.

As mentioned by Jie, we want to select the bayes factor threshold based on the normalized para (```alpha```), while in the result we give credible intervals of the absolute effect size  (```delta```).

shansfolder: The confusion is ```cauchy.stan``` only models ```delta```, so we have to set bf threshold on ```delta``` here, which is another inconsistency. But I am not sure what to do. This is just the easiest fix.

gbordyugov: Shouldn't we better be changing the definition of `alpha` and `delta` in the model files instead?

shansfolder: Yes that's what I did in this PR. Let's discuss in person.

shansfolder:
Updated based on our discussion. @gbordyugov @daryadedik

expan/core/early_stopping.py
```diff
@@ -271,18 +273,28 @@ def bayes_factor(x, y, distribution='normal', num_iters=25000):
...
-            'confidence_interval'   : [{'percentile': p*100, 'value': v} for p, v in zip([p1, p2], interval)],
+            'confidence_interval'   : [{'percentile': p*100, 'value': v} for p, v in zip([p1, p2], credible_interval)],
             'treatment_sample_size' : int(n_x),
             'control_sample_size'   : int(n_y),
             'treatment_mean'        : float(mu_x),
             'control_mean'          : float(mu_y),
             'number_of_iterations'  : num_iters}
 
 
+def get_trace_normalzed_effect_size(distribution, traces):
```
gbordyugov: `normalized`

expan/core/early_stopping.py
```diff
@@ -303,17 +315,22 @@ def bayes_precision(x, y, distribution='normal', posterior_width=0.08, num_iters
         dictionary with statistics
     """
     traces, n_x, n_y, mu_x, mu_y = _bayes_sampling(x, y, distribution=distribution, num_iters=num_iters)
+    trace_normalized_effect_size = get_trace_normalzed_effect_size(distribution, traces)
```
gbordyugov: `normalized`

PR #145, opened 2017-08-07
Documentation updates for Expan 0.6.x. Covers OCTO-1961, OCTO-1970

daryadedik:
1. Changed documentation for Intro.ipynb: removed unnecessary parts (sga, trend, only delta and binning are left), added more comments. 
2. Renaming in comments to csv_fetcher.
3. Added check for correct dictionary structure and fixing broken __str__ for Experiment object.


expan/core/experiment.py
```diff
@@ -73,26 +81,19 @@ def __init__(self, control_variant_name, data, metadata, report_kpi_names=[], de
...
 
-
-
     def get_kpi_by_name_and_variant(self, name, variant):
         return self.data.reset_index().set_index('variant').loc[variant, name]
 
-
     def __str__(self):
-        # res = super(Experiment, self).__str__()
+        res = super(Experiment, self).__str__()
```
gbordyugov: why this constructor call?

expan/core/experiment.py
```diff
@@ -45,6 +46,13 @@ def __init__(self, control_variant_name, data, metadata, report_kpi_names=[], de
         else:
             report_kpi_names_needed = numerical_column_names - experiment_column_names
 
+        # check derived_kpis structure (should have keys namely 'name' and 'formula')
+        for i in derived_kpis:
+            if not isinstance(i, dict):
+                raise TypeError('Derived kpis should be an array of dictionaries')
+            if 'name' not in i or 'formula' not in i:
```
gbordyugov: good idea!

expan/core/experiment.py
daryadedik: commenting that out and assignment res+= produces an error because the variable does not exist. We don't really need the printing, but I decided to leave it, but we need either remove it entirely or just return the string like 
`return '{:d} variants: {}'.format(len(variants),
                                          ', '.join([('*' + k + '*') if (k == self.control_variant_name) else k
                                                    for k in variants]))`

 and remove super because we don't have parenting class anymore.

gbordyugov: then maybe remove the line completely and than turn `+=` into `=` below?

daryadedik: you'r right, that was unnecessary )

expan/core/experiment.py
gbordyugov: btw, what do you think about adding unit tests to cover this new check?

daryadedik: that would be great! will add them shortly!

PR #148, opened 2017-08-08
Release v0.6.1

shansfolder:


gbordyugov:
Saying 'no' would be anyway too late )

PR #149, opened 2017-08-09
Updated intro documentation covering delta methods.

daryadedik:
Updated .ipynb and generated new .html file.

### DIFF
--- a/ExpAn-Intro.html
+++ b/ExpAn-Intro.html
@@ -12146,8 +12146,20 @@ div#notebook {
 <td>Name of the experiment, as known to stakeholders. It can be anything meaningful to you.</td>
 </tr>
 <tr>
-<td><strong>[<code>sources</code>]</strong></td>
+<td><strong><code>[sources]</code></strong></td>
 <td>Names of the data sources used in the preparation of this data.</td>
+</tr>
+<tr>
+<td><strong><code>[experiment_id]</code></strong></td>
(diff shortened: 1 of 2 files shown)


gbordyugov:
Looks great, but that's exactly the problem with this kind of documentation: it changes a lot (styles, formatting, etc.) after small changes.

In the future we have to think of a better alternative.

daryadedik:
AFAIK, in R there is an automated documentation file generation process, you don't need to do formatting you just change fields (name of function, what it does, what parameters it accepts) and create examples and build it (very similar to latex format). Don't know whether there is something like that in Python.

shansfolder:
That's the documentation generated by Sphinx we had in the /docs folder. It can read doc strings and build an api doc, but we are not using it. We should definitely plan for a ticket to have better documentations (not the quick fix we have right now).

shansfolder:
and for the current PR, I have two quick comments:
1. Can you move the section for parameters of deltas up, before the section of interpreting results?
2. Regarding the fields in ```metadata```, I actually meant removing only ```baseline_variant``` (and keep the others as optional, maybe only ```experiment``` as mandatory) because in the example we have more:
```
{'experiment': 'random_data_generation',
 'primary_KPI': 'normal_shifted',
 'source': 'simulated'}
```

daryadedik:
@shansfolder sure, will do the changes.

shansfolder (approved):

PR #150, opened 2017-08-10
OCTO-1971 Add variational inference for early stopping

shansfolder:


tests/tests_core/test_early_stopping.py
```diff
@@ -121,6 +120,13 @@ def test_bayes_factor_with_nan_input(self):
         res= es.bayes_factor(self.rand_s5, self.rand_s6, num_iters=2000)
         self.assertEqual(res['stop'], True)
 
+    def test_variational_inference(self):
+        """
+        Check bayesian sampling using variational bayes.
+        """
+        traces, n_x, n_y, mu_x, mu_y = es._bayes_sampling(self.rand_s1, self.rand_s2, num_iters=2000, inference="variational")
+        self.assertEqual(len(traces['delta']), 1001)
```
gbordyugov: would it be a good idea to check for numerical values too?

shansfolder: The numerical checks are quite unstable for variational inference. I added a few more assertions.

PR #153, opened 2017-08-24
Octo 1930 implement quantile filtering

mkolarek:
Implemented quantile filtering for easier filtering of entities whose KPI values exceed the value of the of the KPI at a given percentile. It is the default (and currently only) method invoked with the filter() method.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -145,16 +145,21 @@ class Experiment(object):
         result['kpis'] = kpis
         return result
 
-    def _quantile_filtering(self, kpis, percentile):
+    def _quantile_filtering(self, kpis, percentile, kind):
         flags = pd.Series(data=[False]*len(self.data))
         for column in self.data[kpis].columns:
             threshold = np.percentile(self.data[column], percentile)
-            flags = flags | self.data[column].apply(
-                lambda x: x >= threshold,
-            )
+            if kind == 'upper':
+                flags = flags | self.data[column].apply(
+                    lambda x: x >= threshold,
+                )
+            elif kind == 'lower':
+                flags = flags | self.data[column].apply(
+                    lambda x: x <= threshold,
+                )
         return flags
 
-    def filter(self, kpis, percentile=99.0):
+    def filter(self, kpis, percentile=99.0, kind='upper'):
         """
         Method that filters out entities whose KPIs exceed the value at a given percentile.
         If any of the KPIs exceeds its threshold the entity is filtered out.
@@ -170,21 +175,26 @@ class Experiment(object):
         # check if provided KPIs are present in the data
         for kpi in kpis:
             if kpi not in self.data.columns:
-                raise KeyError(kpi + ' identifier not present in dataframe columns.')
+                raise KeyError(kpi + ' identifier not present in dataframe columns!')
 
         # check if provided percentile is valid
         if 0.0 < percentile <= 100.0 is False:
             raise ValueError("Percentile value needs to be between 0.0 and 100.0!")
 
+        # check if provided filtering kind is valid
+        if kind not in ['upper', 'lower']:
+            raise ValueError("Kind needs to be either 'upper' or 'lower'!")
+
         # run quantile filtering
-        flags = self._quantile_filtering(kpis=kpis, percentile=percentile)
+        flags = self._quantile_filtering(kpis=kpis, percentile=percentile, kind=kind)
 
         # log which columns were filtered and how many entities were filtered out
         self.metadata['filtered_columns'] = kpis
         self.metadata['filtered_entities_number'] = len(flags[flags == True])
+        self.metadata['filtered_threshold_kind'] = kind
 
         # throw warning if too many entities have been filtered out
-        if (len(flags[flags == True]) / len(self.data)) > 0.02:
+        if (len(flags[flags == True]) / float(len(self.data))) > 0.02:
             warnings.warn('More than 2% of entities have been filtered out, consider adjusting the percentile value.')
 
         self.data = self.data[flags == False]
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -153,21 +153,53 @@ class Experiment(object):
             return True
 
     def _quantile_filtering(self, df, percentile):
-        flags = pd.Series()
+        flags = None
         for column in df.columns:
-            flags = flags | df[column].apply(
-                self._quant_apply,
-                threshold=np.percentile(df[column], percentile)
+            if flags is None:
+                flags = df[column].apply(
+                    self._quant_apply,
+                    threshold=np.percentile(df[column], percentile)
             )
+            else:
+                flags = flags | df[column].apply(
+                    self._quant_apply,
+                    threshold=np.percentile(df[column], percentile)
+                )
         return flags
 
     def filter(self, kpis, percentile=99.0):
+        """
+        Method that filters out entities whose KPIs exceed the value at a given percentile.
+        If any of the KPIs exceeds its threshold the entity is filtered out.
+
+        Args:
+            kpis (list): list of KPI names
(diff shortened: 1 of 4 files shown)


expan/core/experiment.py
```diff
@@ -141,3 +141,62 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+            return False
+        else:
+            return True
+
+    def _quantile_filtering(self, df, percentile):
+        flags = None
+        for column in df.columns:
+            if flags is None:
+                flags = df[column].apply(
+                    self._quant_apply,
```
gbordyugov: Are we planning to use different functions here? otherwise `lambda x, y: x >= y` would suffice and there is no need to declare the static method above

expan/core/experiment.py
```diff
@@ -141,3 +141,62 @@ def delta(self, method='fixed_horizon', **worker_args):
...
         return result
+
+    @staticmethod
+    def _quant_apply(value, threshold):
+        if value < threshold:
+            return False
+        else:
+            return True
+
+    def _quantile_filtering(self, df, percentile):
```
gbordyugov: why pass the `df` parameter and not use the `data` field of the class directly?

expan/core/experiment.py
```diff
@@ -141,3 +141,62 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+        flags = None
+        for column in df.columns:
+            if flags is None:
+                flags = df[column].apply(
```
gbordyugov: can we initialize `flags` with smth like `flags = [False]*len(df[column])` and drop the `if` condition altogether (since the `then` and the `else` arms of the conditional are nearly identical)?

expan/core/experiment.py
```diff
@@ -144,3 +144,62 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+                    self._quant_apply,
+                    threshold=np.percentile(df[column], percentile)
+            )
+            else:
+                flags = flags | df[column].apply(
+                    self._quant_apply,
+                    threshold=np.percentile(df[column], percentile)
+                )
```
gbordyugov: or even smth like for the whole function

```
def _quantile_filtering(self, df, percentile):
  from functools import reduce

  def update_flags(flags, column):
      return flags | df[column].apply(self._quant_apply, threshold=np.percentile(df[column], percentile)

  init_flags = [False]*len(df[columns[0]])
  return reduce(update_flags, columns, init_flags)
```

expan/core/experiment.py
```diff
@@ -144,3 +144,62 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+        metadata = {
+            'filtered_columns': kpis,
+            'filtered_entities_number': len(flags[flags == True])
+        }
+
+        # throw warning if too many entities have been filtered out
+        if (len(flags[flags == True]) / len(self.data)) > 0.02:
+            warnings.warn('More than 2% of entities have been filtered out, consider adjusting the percentile value.')
+
+        self.metadata = metadata
```
gbordyugov: We have to work out a convention about metadata: should it be a list or a dict with all possible pieces of metadata

gbordyugov (commented):
Just a few comments

expan/core/experiment.py
mkolarek: agreed, this is just the simplest form I could think of

expan/core/experiment.py
mkolarek: this could be simplified, true

expan/core/experiment.py
mkolarek: initially I didn't want to modify the existing data, but ultimately I decided to keep it simple... this is a leftover of the first approach, will fix

expan/core/experiment.py
mkolarek: will look into this

tests/tests_core/test_experiment.py
```diff
@@ -241,6 +241,47 @@ def test_bayes_precision_delta_derived_kpis(self):
...
+            kpis=[
+                'normal_same',
+                'normal_shifted',
+                'normal_shifted_by_feature',
+                'normal_unequal_variance'
+            ]
+        )
+        self.assertEqual(len(self.data) - len(exp.data), exp.metadata['filtered_entities_number'])
+
+    def test_quantile_filtering_2(self):
```
shansfolder: I would give it a more meaningful name. Also for the other quantile filtering tests.

expan/core/experiment.py
```diff
@@ -144,3 +144,62 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+        if 0.0 < percentile <= 100.0 is False:
+            raise ValueError("Percentile value needs to be between 0.0 and 100.0!")
+
+        flags = self._quantile_filtering(df=self.data[kpis], percentile=percentile)
+
+        # log which columns were filtered and how many entities were filtered out
+        metadata = {
+            'filtered_columns': kpis,
+            'filtered_entities_number': len(flags[flags == True])
+        }
```
shansfolder: We already have ```self.metadata``` when the experiment object is initialized, so I would read ```self.metadata first``` and then add these two fields to it.

shansfolder (changes requested):

tests/tests_core/test_experiment.py
mkolarek: true, names improved

shansfolder (approved):

gbordyugov:
Marko btw do you compare against just one percentile supposing that the outliers have always larger KPI values?

mkolarek:
@gbordyugov yes, I always assume that the outlier values are higher than the percentile value... this is a good point, perhaps I can add an additional parameter that indicates what kind of threshold the specified percentile should create

expan/core/experiment.py
```diff
@@ -144,3 +144,47 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+
+        # run quantile filtering
+        flags = self._quantile_filtering(kpis=kpis, percentile=percentile)
+
+        # log which columns were filtered and how many entities were filtered out
+        self.metadata['filtered_columns'] = kpis
+        self.metadata['filtered_entities_number'] = len(flags[flags == True])
+
+        # throw warning if too many entities have been filtered out
+        if (len(flags[flags == True]) / len(self.data)) > 0.02:
```
gbordyugov: In Python 2, an `int` divided by another `int` results in an `int`

expan/core/experiment.py
```diff
@@ -144,3 +144,62 @@ def delta(self, method='fixed_horizon', **worker_args):
 
         result['kpis'] = kpis
         return result
+
+    @staticmethod
+    def _quant_apply(value, threshold):
+        if value < threshold:
+            return False
+        else:
+            return True
```
daryadedik: i suppose, can be written simpler:

`def _quant_apply(value, threshold):
      return value > threshold`

mkolarek:
@gbordyugov I've added a parameter for specifying the kind of threshold used for the filtering

expan/core/experiment.py
```diff
@@ -144,3 +144,57 @@ def delta(self, method='fixed_horizon', **worker_args):
...
+
+    def filter(self, kpis, percentile=99.0, kind='upper'):
+        """
+        Method that filters out entities whose KPIs exceed the value at a given percentile.
+        If any of the KPIs exceeds its threshold the entity is filtered out.
+
+        Args:
+            kpis (list): list of KPI names
+            percentile (float): percentile considered as threshold
+
```
shansfolder: I would document what does ```kind``` mean.
Besides, the word ```kind``` sounds kind of inaccurate. Maybe rename to ```filter_out_type```?

mkolarek: good catch, forgot to document that. will rename it as well.

PR #155, opened 2017-08-29
OCTO-1920, OCTO-1968, OCTO-1969 Refactor binning

shansfolder:
This PR contains
* skeleton of the binning refactoring
* numerical binning
* categorical binning
* new unit tests
* updated ExpAn-Intro notebook
* ExpAn-Intro html file


expan/core/binning.py
```diff
@@ -8,725 +6,243 @@
...
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
+
+class NumericalRepresenation(object):
```
daryadedik: small typos in the names; Numerical**Represenation** and Numerical**Represenation**.

shansfolder: thanks

expan/core/binning.py
```diff
@@ -8,725 +6,248 @@
...
+    def __init__(self, bin_type, *repr_args):
+        '''
+        Constructor for a bin object.
+        :param id: identifier (e.g. bin number) of the bin
+        :param bin_type: "numerical" or "categorical"
+        :param repr_args: arguments to represent this bin. 
+                          args for numerical bin includes lower, upper, lower_closed, upper_closed
+                          args for categorical bin includes a list of values
+        '''
+        if bin_type == "numerical" and len(repr_args) != 4:
```
igusher: These conditions feels like adding two derived classes might be a good idea to consider.
For instance:
`class NumericalBin(Bin):
def __init__(self, lower, upper, lower_closed, upper_closed)
class CategoricalBin(Bin):
def __init__(self, categories)`

shansfolder: do you mean use ```class NumericalBin(Bin)``` and get rid of ```NumericalRepresentation```?

igusher: Well, yes, maybe in this way, like you say.
The point is, as far as I see the only responsibility of Bin class is to receive some parameters, and decide which binning it is.
But one can solve it on the level of class design.

mkolarek (approved):

expan/core/binning.py
igusher: Feel free to postpone it, it is a stylistic comment, we can refactor it later.
[PS. However, "there is no time for refactoring"]

daryadedik (approved):

PR #157, opened 2017-08-31
categorical binning backend

gbordyugov:
After having read about tricky heuristics to solve the NP-hard problem of binning, I decided to go with a simplest greedy algorithm: sort the items in the decreasing order of their weight and then assign them to the current lightest bin.

This is just a backend of the binning, i.e. a function that return a list of pairs `(weight, list_of_items)`. Need to wrap it into classes, I guess.

### DIFF
--- a/expan/core/categorical_binning.py
+++ b/expan/core/categorical_binning.py
@@ -1,0 +1,37 @@
+from heapq import heapify, heappush, heappop
+
+def categorical_binning(items, n_bins):
+    """ Performs greedy (non-optimal) binning
+    :param items: list of items to bin according to their frequencies
+    :return: list of bins [(total_weight, [items])]
+    """
+
+    # count items
+    weights = {}
+    for item in items:
+        if not item in weights:
+            weights[item] = 1
+        else:
+            weights[item] += 1
+
+    # we need items sorted in decreasing order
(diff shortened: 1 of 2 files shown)


expan/core/categorical_binning.py
```diff
@@ -0,0 +1,36 @@
...
+        if not item in weights:
+            weights[item] = 1
+        else:
+            weights[item] += 1
+
+    # we need items sorted in decreasing order
+    pairs = sorted([(weight, [item]) for (item, weight) in weights.items()], reverse=True)
+
+    # take max(n_bins, len(pairs)) haviest pairs as initial bins
+    bins = [(weight, labels) for (_, (weight, labels)) in zip(range(n_bins), pairs)]
```
igusher: is it the same as pairs[:n_bins] ?

gbordyugov: not quite.

`pairs` might happen to be have fewer than `n_bins` elements (`zip` would terminate at the shortest of two)

igusher:
:+1:

daryadedik:
👍

daryadedik (approved):

shansfolder:
integrated into PR #155

PR #159, opened 2017-09-07
OCTO-1933 Subgroup analysis

shansfolder:
This PR contains
* add statistical power to result.
* implement sga
* cleanup / fix some old tests


shansfolder:
Ready for review now. :)

mkolarek (approved):

expan/core/binning.py
```diff
@@ -95,24 +96,28 @@ def __eq__(self, other):
...
 
         if self.lower_closed:
-            filter_lower = (data >= self.lower)
+            filter_lower = (data_subgroup_dimension >= self.lower)
         else:
-            filter_lower = (data > self.lower)
+            filter_lower = (data_subgroup_dimension > self.lower)
         if self.upper_closed:
-            filter_upper = (data <= self.upper)
+            filter_upper = (data_subgroup_dimension <= self.upper)
```
daryadedik: just to clarify: is data_subgroup_dimension if the dimension of one subgroup? does it mean if there are 3 kpis then the data_subgroup_dimension will be 3?

daryadedik:
👍

daryadedik (approved):

expan/core/binning.py
shansfolder: It will only be one column (the given feature name) of the data. 
-> then we compare this column with the defined bin value to find out what belongs to the subgroup 
-> results in a column of booleans
-> then we use these booleans to filter the complete dataframe again (here you will have a dataframe with kpi1, kpi2, and all other columns but only with rows belongs this subgroup)

daryadedik: clear! thanks!

shansfolder: I now renamed it to ```data_feature_column```.

Also, I renamed everything I called ```dimension``` to ```feature```, to be consistent with analysis-service in the backend.

PR #161, opened 2017-09-20
OCTO-2088: Implement multiple testing correction in ExpAn

daryadedik:
Bonferroni correction for confidence intervals. For more explanations please refer to OCTO-2088.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -24,10 +24,10 @@ class Experiment(object):
     """
     def __init__(self, control_variant_name, data, metadata, report_kpi_names=None, derived_kpis=None):
         report_kpi_names = report_kpi_names or []
-        derived_kpis  = derived_kpis or []
+        derived_kpis = derived_kpis or []
 
         experiment_column_names = set(['entity', 'variant'])
-        numerical_column_names  = set(get_column_names_by_type(data, np.number))
+        numerical_column_names = set(get_column_names_by_type(data, np.number))
 
         if type(report_kpi_names) is str:
             report_kpi_names = [report_kpi_names]
@@ -77,10 +77,8 @@ class Experiment(object):
             self.data.loc[:, name] = eval(re.sub(kpi_name_pattern, r'self.data.\1.astype(float)', formula))
             self.reference_kpis[name] = re.sub(kpi_name_pattern + '/', '', formula)
 
-
     def get_kpi_by_name_and_variant(self, data, name, variant):
         return data.reset_index().set_index('variant').loc[variant, name]
-
 
     def __str__(self):
         variants = self.variant_names
@@ -88,7 +86,6 @@ class Experiment(object):
         return 'Experiment "{:s}" with {:d} derived kpis, {:d} report kpis, {:d} entities and {:d} variants: {}'.format(
             self.metadata['experiment'], len(self.derived_kpis), len(self.report_kpi_names), len(self.data),
             len(variants), ', '.join([('*' + k + '*') if (k == self.control_variant_name) else k for k in variants]))
-
 
     def _get_weights(self, data, kpi, variant):
         if kpi not in self.reference_kpis:
@@ -99,10 +96,8 @@ class Experiment(object):
         non_zeros      = len(x) - zeros_and_nans
         return non_zeros/np.nansum(x) * x
 
-
     def delta(self, method='fixed_horizon', **worker_args):
         return self._delta(method=method, data=self.data, **worker_args)
-
 
     def _delta(self, method, data, **worker_args):
         worker_table = {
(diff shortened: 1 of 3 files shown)


daryadedik:
unit tests will be added shortly

expan/core/experiment.py
```diff
@@ -141,14 +134,13 @@ def _delta(self, method, data, **worker_args):
                     power = statx.compute_statistical_power(treatment_data, control_data)
                     statistics['statistical_power'] = power
                 if len(w):
-                    result['warnings'].append('kpi: ' + kpi + ', variant: '+ variant + ': ' + str(w[-1].message))
-                res_kpi['variants'].append({'name': variant, 'delta_statistics' : statistics})
+                    result['warnings'].append('kpi: {}, variant: {}: {}'.format(kpi, variant, w[-1].message))
```
gbordyugov, line 137: do we have a standard on that?

expan/core/statistics.py
```diff
@@ -89,8 +91,10 @@ def delta(x, y, assume_normal=True, percentiles=[2.5, 97.5],
         mu = _delta_mean(_x, _y)
         # Computing the confidence intervals
         if assume_normal:
-            c_i = normal_sample_difference(x=_x, y=_y, percentiles=percentiles,
-                                           relative=relative)
+            if num_tests:
+                percentiles = [x/num_tests if x < 50.0
+                               else 100-(100-x)/num_tests if x > 50.0 else x for x in percentiles]
```
gbordyugov: in Python 2, dividing an `int` by an `int` always produces an `int`, please make sure that no integer division happens here

expan/core/experiment.py
daryadedik: this is just simpler version of the concatentation. Here you don't need to do type casting implicitly.

gbordyugov: Oh, you're right, I thought you were constructing a `dict` instead of a `str`, sorry

expan/core/statistics.py
daryadedik: yes, makes sense, thank you!

mkolarek: as a general point we could discuss dropping Python 2 support all together

expan/core/statistics.py
```diff
@@ -4,24 +4,25 @@
...
 
+
 def _delta_mean(x, y):
     """Implemented as function to allow calling from bootstrap. """
     return np.nanmean(x) - np.nanmean(y)
 
 
 def make_delta(assume_normal=True, percentiles=[2.5, 97.5],
-               min_observations=20, nruns=10000, relative=False):
+               min_observations=20, nruns=10000, relative=False, num_tests=0):
```
gbordyugov: `num_tests=0`?

gbordyugov (changes requested):

expan/core/statistics.py
```diff
@@ -89,8 +91,9 @@ def delta(x, y, assume_normal=True, percentiles=[2.5, 97.5],
         mu = _delta_mean(_x, _y)
         # Computing the confidence intervals
         if assume_normal:
-            c_i = normal_sample_difference(x=_x, y=_y, percentiles=percentiles,
-                                           relative=relative)
+            percentiles = [float(p)/num_tests if p < 50.0
+                           else 100-(100-float(p))/num_tests if p > 50.0 else p for p in percentiles]
+            c_i = normal_sample_difference(x=_x, y=_y, percentiles=percentiles, relative=relative)
         else:
             c_i, _ = bootstrap(x=_x, y=_y, percentiles=percentiles, nruns=nruns,
```
gbordyugov: and what about the non-normal distributions (I don't know either)?

daryadedik: good question, but I am not sure about the correction for the bootstrap. Will investigate that.

gbordyugov (approved):
great up to the open question of correcting the bootstrap results

expan/core/statistics.py
daryadedik: so, basically the question here is not about to correct for non-normal distribution or not (correction as it is does not depend on the distribution of the data), but whether to correct for the bootstrap or not. In our scenario it make sense to make the same correction of CIs as for the hypothesis testing, since here we report corrected confidence intervals and, basically, they should be approximately similar to CIs reported by hypothesis testing. Bootstrap is not a hypothesis testing though, but in our case it may be considered as another way of representing the data. Here we correct more in terms of the decision on how to select a confidence interval: the more kpis we observe the higher the chance that one kpi will give different results than others.  But if you have other ideas, I would be happy to discuss.

expan/core/experiment.py
daryadedik: no problem!

daryadedik:
I merged this pull request due to no other comments.

PR #164, opened 2017-10-19
OCTO-2181: Implement over time analysis. Time-based SGA

daryadedik:
Notes:
1. Binning is done per day
2. 'date' column is expected
3. Added variational inference use and short documentation update


gbordyugov:
will review today!

expan/core/experiment.py
```diff
@@ -190,36 +190,51 @@ def filter(self, kpis, percentile=99.0, threshold_type='upper'):
...
-            for bin in feature_name_to_bins[feature]:
-                subgroup = {'dimension': feature,
-                            'segment': str(bin.representation)}
-                subgroup_data = bin.apply(self.data, feature)
-                subgroup_res = self._delta(method='fixed_horizon', data=subgroup_data,
-                                           num_tests=len(self.report_kpi_names))
-                subgroup['result'] = subgroup_res
-                subgroups.append(subgroup)
-
+        if feature_name_to_bins:
```
gbordyugov: Since `feature_name_to_bins` has no default `None` value, I'm wondering why this test?

expan/core/experiment.py
```diff
@@ -190,36 +190,51 @@ def filter(self, kpis, percentile=99.0, threshold_type='upper'):
 
         self.data = self.data[flags == False]
 
-    def sga(self, feature_name_to_bins):
+    def sga(self, feature_name_to_bins, time_interval=None):
```
gbordyugov: Shouldn't time binning be a special case of binning in general? If so, the information about it should be contained in `feature_name_to_bins`

expan/core/experiment.py
```diff
@@ -190,36 +190,51 @@ def filter(self, kpis, percentile=99.0, threshold_type='upper'):
...
+                for bin in feature_name_to_bins[feature]:
+                    subgroup = {'dimension': feature,
+                                'segment': str(bin.representation)}
+                    subgroup_data = bin.apply(self.data, feature)
+                    subgroup_res = self._delta(method='fixed_horizon', data=subgroup_data,
+                                               num_tests=len(self.report_kpi_names))
+                    subgroup['result'] = subgroup_res
+                    subgroups.append(subgroup)
+
+        if time_interval:
```
gbordyugov: Here again - wouldn't it be slightly more general not to try to make a special case of binning in time?

expan/core/experiment.py
daryadedik: afair we discussed with the product the possibility to choose the time interval on which to run the experiment. So, I consider that as one more segment in the sga. Doing some random bins on data can be not very useful for the stakeholder, because the interval he is interested in may not be covered with the bins. Probably by day binning could be interesting, but that can be done just by grouping by day. I can discuss that with Adrien.

gbordyugov: One potential way to go would be to map the time column into an additional column with Unix time (https://en.wikipedia.org/wiki/Unix_time) and, Unix time being just an `int` of seconds, apply numerical binning on that column.

daryadedik: sure, that will be binning by time since treatment, which can be added and is useful I think. I can add that. But I am not sure it is good to create additional column in data in Expan, but probably do that in Expan service.

gbordyugov: Another idea would be to create an additional column with date only (no hours:minutes:seconds) an group the data frame by that column...

expan/core/experiment.py
daryadedik: sorry, forgot to remove that, just left from my previous different implementation.

expan/core/experiment.py
daryadedik: correct me if I am wrong, but date doesn't have timestamp currently, the format is 'YYYYDDMM' or 'YYYY-DD-MM', so, the categorical binning already can make bins on that column. But I can add unit tests for checking that.

expan/core/experiment.py
```diff
@@ -217,9 +219,24 @@ def sga(self, feature_name_to_bins):
...
+            if 'date' in self.data:
+                subgroup_data = self.data[(start <= self.data.date) & (self.data.date <= end)]
+            else:
+                raise KeyError("No date column is provided in data for time-based SGA!")
+
+            if subgroup_data is not None:
+                subgroups.append({'dimension': 'date',
+                                  'segment': time_interval,
+                                  'result': self._delta(method='fixed_horizon', data=subgroup_data,
+                                                        num_tests=len(self.report_kpi_names))})
```
shansfolder: If we provide both ```feature_name_to_bins``` and ```time_interval```, it will not do the filtering on dates first and then do the subgroup based on bins, right? but time interval is just an additional bin, which will not influence data for other bins.

The signature is a little bit confusing to use. If this is intended, could you also document it in the corresponding ```readthedocs``` page?

shansfolder (changes requested):

expan/core/experiment.py
shansfolder: oh and btw, I added a check: if subgroup data is None based on the filtering, we don't do ```_delta```(because it will return an error), but skip this subgroup and continue the next. Same applies for time interval based subgroup.

daryadedik: @shansfolder I can move the subsetting data by time_interval before subgroups, so that we also can first filter on dates and then do subgroups. If only time_interval is set, there will be only 1 subgroup by date. Is that what you meant?

shansfolder:
@daryadedik is this PR in the state for review now?

shansfolder (approved):

daryadedik:
Update: added exposure for variational inference. Wanted to make as another pull request. Sorry for messy PR.

expan/core/binning.py
```diff
@@ -17,14 +17,14 @@ class Bin(object):
     __hash__ = None
 
     def __init__(self, bin_type, *repr_args):
-        '''
+        """
```
gbordyugov, line 20: is it kind of misalignment between default quotes in different IDEs or was it a real mistake using single ticks?

daryadedik: just for style consistency with other parts of code, nothing special.

shansfolder:
👍

PR #170, opened 2017-10-24
Add some parameter checks

shansfolder:
As we discussed in the meeting.

### DIFF
--- a/expan/core/binning.py
+++ b/expan/core/binning.py
@@ -128,14 +128,15 @@ class CategoricalRepresentation(object):
     __hash__ = None
 
     def __init__(self, categories):
-        '''
+        """
         Constructor for representation of a categorical bin.
         :param categories: list of categorical values that belong to this bin
-        '''
-        if type(categories) is list:
+        """
+        try:
             categories = set(categories)
-        if type(categories) is not set:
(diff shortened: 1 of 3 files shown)


expan/core/binning.py
```diff
@@ -132,12 +132,14 @@ def __init__(self, categories):
         Constructor for representation of a categorical bin.
         :param categories: list of categorical values that belong to this bin
         '''
-        if type(categories) is not list:
-            raise ValueError("args for categorical bin is a list of categorical values.")
+        if type(categories) is list:
```
gbordyugov: what happens if it isn't list?

we could try just to cast to set and catch an exception

shansfolder: It is in the next lines. If it's neither list nor set, we will raise an error.

gbordyugov: Yeah, I'm just saying that
```
categories = set(categories)
```
would do the same job as four lines (135 through 138). Please correct me if I'm wrong.

mkolarek (approved):

expan/core/binning.py
shansfolder: Right..Done!

PR #173, opened 2017-11-22
Fix two-sided alpha value in power analysis

shansfolder:
Aligned with Tomas


gbordyugov:
could you please provide a little bit more background about this change?

shansfolder:
As discussed, when we derive the formula of power analysis, it is based on the meeting point of false positive and false negative (beta). In our case, alpha is always two-sided, so the area of false positive that overlaps with beta is alpha/2.

(It's pretty hard to explain in plain words. If anyone has doubts, I can explain in person by drawing)

daryadedik:
👍

PR #174, opened 2017-11-24
AXO-35 implemented estimate_sample_size() for estimating sample size …

mkolarek:
…based on sample mean and variance given an MDE, number of variants and variant split ratio

### DIFF
--- a/expan/core/statistics.py
+++ b/expan/core/statistics.py
@@ -134,6 +134,37 @@ def sample_size(x):
     return len(x) - x_nan
 
 
+def estimate_sample_size(x, mde, n, r, alpha=0.05, beta=0.2):
+    """
+    Estimates sample size based on sample mean and variance given MDE (Minimum Detectable effect), number of variants and variant split ratio
+
+    Args:
+        x (pd.Series or pd.DataFrame): sample to base estimation on
+        mde (float): minimum detectable effect
+        n (int): number of variants
+        r (float): variant split ratio
+        alpha (float): significance level
+        beta (float): type II error
+
+    Returns:
+        float or pd.Series: estimated sample size
+
+    """
+    if not isinstance(x, pd.Series) and not isinstance(x, pd.DataFrame):
+        raise TypeError("Sample x needs to be either Series or DataFrame.")
+
+    if not isinstance(n, int):
+        raise TypeError("Number of variants needs to be an integer.")
+
+    if n <= 0:
+        raise ValueError("Number of variants needs to be higher than 0.")
+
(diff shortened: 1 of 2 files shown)


expan/core/statistics.py
```diff
@@ -134,6 +134,37 @@ def sample_size(x):
...
+    if not isinstance(n, int):
+        raise TypeError("Number of variants needs to be an integer.")
+
+    if n <= 0:
+        raise ValueError("Number of variants needs to be higher than 0.")
+
+    if r <= 0:
+        raise ValueError("Variant split ratio needs to be higher than 0.")
+
+    return n * ((stats.norm.ppf(1 - alpha) - stats.norm.ppf(beta)) ** 2 * (x.var() / (x.mean() ** 2)) * (1 + (1 / float(r)))) / (mde ** 2)
```
shansfolder: This should be alpha/2. See my PR: https://github.com/zalando/expan/pull/173

expan/core/statistics.py
```diff
@@ -134,6 +134,37 @@ def sample_size(x):
     return len(x) - x_nan
 
 
+def estimate_sample_size(x, mde, n, r, alpha=0.05, beta=0.2):
+    """
+    Estimates sample size based on sample mean and variance given MDE (Minimum Detectable effect), number of variants and variant split ratio
+
+    Args:
+        x (pd.Series or pd.DataFrame): sample to base estimation on
```
gbordyugov, line 142: just curious - why not `numpy.array`?

expan/core/statistics.py
```diff
@@ -134,6 +134,37 @@ def sample_size(x):
     return len(x) - x_nan
 
 
+def estimate_sample_size(x, mde, n, r, alpha=0.05, beta=0.2):
```
shansfolder, line 137: It's an isolated method for now and we don't expose this to result yet, right?

expan/core/statistics.py
mkolarek: could be an array, there's a nice benefit of allowing it to be a DataFrame since we can calculate the estimate for multiple columns simultaneously

gbordyugov: @mkolarek thanks, I see!

expan/core/statistics.py
mkolarek: yes, it has to be called separately

expan/core/statistics.py
mkolarek: I see, thanks!

shansfolder (approved):

PR #184, opened 2018-01-19
Axo-91 bug fix sga

shansfolder:
Fix the bug that sga will raise error when there is only one variant in the subgroup.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -226,7 +226,7 @@ class Experiment(object):
                             'segment': str(bin.representation)}
                 subgroup_data = bin(self.data, feature)
 
-                if subgroup_data is None:
+                if self._isNotValidForAnalysis(subgroup_data):
                     continue
 
                 subgroup_res = self._delta(method='fixed_horizon', data=subgroup_data,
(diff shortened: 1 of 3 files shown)


expan/core/experiment.py
```diff
@@ -236,6 +236,21 @@ def sga(self, feature_name_to_bins, multi_test_correction=False):
 
         return subgroups
 
+    def _isNotValidForAnalysis(self, df):
```
gbordyugov: Just a minor thing: why not `_isValidForAnalysis`?

One more general thing: I think that having a lot of methods that work on external dataframes goes a bit against the idea of encapsulation. The potential problem here is `df` is only particularly verified and all the checks that we run in the c'tor of `Experiment` are missing (please don't duplicate that code though :~)

shansfolder: ```_isValidForAnalysis``` is a bit better. True.

Well, those general checks will still be run when we create an ```Experiment``` object regardlessly, so they are already verified. This method in this PR is an additional check when we are in ```sga```.

gbordyugov: What I was saying is that no-one prevents you from calling `_isNotValidForAnalysis` with a totally random, `Experiment`-instance unrelated dataframe

gbordyugov: But I guess it's a consequence of our design decision to split dataframes for SGA. Alternatively, we could spawn daughter instances of `Experiment` for each SGA sub-test.

shansfolder: Got it. I recall the discussion now. We should keep track of technical debts.

daryadedik:
looks good! thanks!

PR #185, opened 2018-01-24
AXO-103 include variance in delta / group-sequential reports

gbordyugov:
AXO-103

include variance in delta / group-sequential reports


shansfolder (approved):
Please update the doc too.

daryadedik (approved):

PR #188, opened 2018-01-30
fixing sample size estimation

gbordyugov:
We still don't know how to handle multiple variants.

The old version had both `n` as `r` as parameters.

### DIFF
--- a/expan/core/statistics.py
+++ b/expan/core/statistics.py
@@ -165,7 +165,10 @@ def estimate_sample_size(x, mde, n, r, alpha=0.05, beta=0.2):
     if r <= 0:
         raise ValueError("Variant split ratio needs to be higher than 0.")
 
-    return n * ((stats.norm.ppf(1 - alpha / 2.) - stats.norm.ppf(beta)) ** 2 * (x.var() / (x.mean() ** 2)) * (1 + (1 / float(r)))) / (mde ** 2)
+    ppf = stats.norm.ppf
+    c1 = (ppf(1.0 - alpha/2.0) - ppf(beta))**2
+    c2 = (1.0 + r) * c1 * (1.0 + 1.0 / r)
+    return c2 * x.var() / (mde * x.mean())**2
 
 
 def chi_square(x, y, min_counts=5):
--- a/expan/core/statistics.py
+++ b/expan/core/statistics.py
@@ -137,14 +137,13 @@ def sample_size(x):
     return len(x) - x_nan
 
 
-def estimate_sample_size(x, mde, n, r, alpha=0.05, beta=0.2):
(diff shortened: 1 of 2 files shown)


gbordyugov:
Ok, we can run sample size estimation against control for all variants (except the control itself), giving us a sample size estimation for each variant and then take the largest of them.

What do you think?

expan/core/statistics.py
```diff
@@ -165,7 +165,10 @@ def estimate_sample_size(x, mde, n, r, alpha=0.05, beta=0.2):
     if r <= 0:
         raise ValueError("Variant split ratio needs to be higher than 0.")
 
-    return n * ((stats.norm.ppf(1 - alpha / 2.) - stats.norm.ppf(beta)) ** 2 * (x.var() / (x.mean() ** 2)) * (1 + (1 / float(r)))) / (mde ** 2)
+    ppf = stats.norm.ppf
+    c1 = (ppf(1.0 - alpha/2.0) - ppf(beta))**2
+    c2 = (1.0 + r) * c1 * (1.0 + 1.0 / r)
+    return c2 * x.var() / (mde * x.mean())**2
```
shansfolder, line 164: Thanks for the fix. 
Why do we need ```x.mean```? I couldn't relate this part in the power analysis formula.

shansfolder (commented):
I think we don't need `n`: number of variants in this method. Besides, this method is not used in Expan as well as not in the doc.
 
I can't remember when and why we have this method in the first place.

expan/core/statistics.py
gbordyugov: here, `mde` is a relative effect size, hence `mde * mean` is the corresponding absolute effect size

gbordyugov:
@shansfolder this method is called from our `ExpanWeb` service and the data is currently saved along other experiment results.

shansfolder:
@gbordyugov I see. Then I would suggest to discuss the signature (e.g. `n` is not used) and properly document this method.

One more question: if there are more variants, do we run this method multiple times and take the sum?

gbordyugov:
@shansfolder will do

as to your second questions, see my comment above — the one ending with "What do you think"?

jbao:
@gbordyugov I like your idea of doing the iterative sample size estimation for multivariate tests, maybe it's also possible to do the power calculation in terms of the Chi-squared statistic (instead of t- or z-statistic), not quite sure though.

PS: really impressed by the quick hot-fix;-)

jbao (approved):

expan/core/statistics.py
jbao: I agree more docs would also be helpful, took me a while to figure out;-)

PR #193, opened 2018-02-19
Results data structure

shansfolder:


gbordyugov:
Woah guys, great job, I'll review it tomorrow — looking forward to diving into the code!

expan/core/statistical_test.py
```diff
@@ -0,0 +1,70 @@
...
+    :param variants: variant column name and their values
+    :type  variants: Variants
+    :param formula: formula of the kpi, if this is a derived kpi
+    :type  formula: str
+    """
+    def __init__(self, kpi_name, features, variants, formula=None):
+        self.kpi_name  = kpi_name
+        self.features  = features
+        self.variants  = variants
+        self.formula   = formula
```
shansfolder: The design of derived kpi is one place that might need discussion.

expan/core/results.py
```diff
@@ -0,0 +1,110 @@
...
+        self.confidence_interval = c_i
+
+
+# --------- Below are the data structure of test results --------- #
+class StatisticalTestResult(object):
+    """ This class holds the results of a single statistical test.
+    :param test: information about the statistical test
+    :type  test: StatisticalTest
+    :param result: result of this statistical test
+    :type  result: BaseTestStatistics or its subclasses or CorrectedTestStatistics  #TODO: better approach?
```
shansfolder, line 96: Here is another place I found our design can be improved.

expan/core/results.py
```diff
@@ -0,0 +1,110 @@
...
+    original_test_statistics and corrected_test_statistics should have the same type.
+    :param original_test_statistics: test result before correction
+    :type  original_test_statistics: SimpleTestStatistics or EarlyStoppingTestStatistics
+    :param corrected_test_statistics: test result after correction
+    :type  corrected_test_statistics: SimpleTestStatistics or EarlyStoppingTestStatistics
+    """
+    def __init__(self, original_test_statistics, corrected_test_statistics):
+        type1 = type(original_test_statistics)
+        type2 = type(corrected_test_statistics)
+        if type1 != type2:
```
gbordyugov, line 72: we could also double check if those are instances of subclasses of `BaseTestStatistics`, i.e. by

`isinstance(original_test_statistics, BaseTestStatistics)`

expan/core/results.py
gbordyugov: @shansfolder yes, I feel it too, but don't have an immediate solution now.

I think it's fine for the time being, since we can change the implementation details later and leave the interface untouched by introducing/redefining `test` and `results` properties.

expan/core/statistical_test.py
gbordyugov: An explicit class for KPIs and derived one as a subclass?

expan/core/results.py
shansfolder: good point!

expan/core/statistical_test.py
shansfolder: We've thought about it too. Not sure if this might become additional complexity?

gbordyugov: Maybe we should first think of an interface of the `KPI` functionality

For Expan, KPIs are just columns in the data frame, don't you think? Or is there any additional functionality?

shansfolder: yeah. for derived kpis, it is used for creating a new column and store the referencing column.

expan/core/results.py
```diff
@@ -67,6 +67,10 @@ class CorrectedTestStatistics(object):
     :type  corrected_test_statistics: SimpleTestStatistics or EarlyStoppingTestStatistics
     """
     def __init__(self, original_test_statistics, corrected_test_statistics):
+        if not isinstance(original_test_statistics, BaseTestStatistics):
+            raise RuntimeError("original_test_statistics should be instance of BaseTestStatistics or its subclass")
+        if not isinstance(corrected_test_statistics, BaseTestStatistics):
```
gbordyugov: one could first check if both types are the same, and after that check that one of them `isinstance`

shansfolder:
@gbordyugov done

gbordyugov (approved):
great job!

PR #194, opened 2018-02-21
Project Headache

shansfolder:
Please don't merge yet. Let's merge this branch until all related refactoring is done and tested.


expan/core/experiment.py
```diff
@@ -1,280 +1,232 @@
...
+        :param **worker_args: additional arguments for the analysis method
+        :return: statistical result of the test
+        :rtype: StatisticalTestResult
+        """
+        if not isinstance(test, StatisticalTest):
+            raise TypeError("Statistical test should be of type StatisticalTest.")
+
+        if 'entity' not in self.data.columns():
+            raise RuntimeError("There is no 'entity' column in the data.")
+        if self.data.entity.duplicated().any():
```
gbordyugov: I think that it used to be a concern of Expan to filter out those duplicates. Am I wrong?

expan/core/experiment.py
```diff
@@ -1,280 +1,232 @@
...
+            raise ValueError('Entities in data should be unique.')
+
+        if test.variants.variant_column_name not in self.data.columns():
+            raise RuntimeError("There is no '{}' column in the data.".format(test.variants.variant_column_name))
+        if test.variants.treatment_name not in np.unique(self.data[test.variants.variant_column_name]):
```
gbordyugov: Why not use `set()` instead of `np.unique()`? Is the latter substantially faster than the former?

expan/core/experiment.py
```diff
@@ -1,280 +1,232 @@
...
+            raise RuntimeError("There is no treatment with the name '{}' in the data.".format(test.variants.treatment_name))
+        if test.variants.control_name not in np.unique(self.data[test.variants.variant_column_name]):
+            raise RuntimeError("There is no control with the name '{}' in the data.".format(test.variants.control_name))
+
+        if not isinstance(test.features, list):
```
gbordyugov: This should be checked in the constructor of 'StatisticalTest` and not here, I believe.

expan/core/experiment.py
```diff
@@ -1,280 +1,232 @@
...
+            raise TypeError("Features should be a list.")
+        if not all(isinstance(n, FeatureFilter) for n in test.features):
```
gbordyugov: This check should also go into the constructor of the test, if I'm not mistaken.

tests/tests_core/test_statistical_test.py
```diff
@@ -0,0 +1,70 @@
...
+        multi_test_suite = StatisticalTestSuite(tests, "bh")
+
+        self.assertEqual(multi_test_suite.size, 4)
+        self.assertEqual(multi_test_suite.correction_method, "bh")
+
+    def test_make_derived_kpi(self):
+        numerator = "normal_same"
+        denominator = "normal_shifted"
+        derived_kpi_name = "derived_kpi_one"
+        DerivedKPI(derived_kpi_name, numerator, denominator).make_derived_kpi(self.data)
```
gbordyugov, line 61: `make_derived_kpi()` should be ensured to perform its deeds only once, either here or within the `DerivedKPI.make_derived_kpi()` method

expan/core/experiment.py
```diff
@@ -1,280 +1,232 @@
...
+        # apply feature filter to data
+        for feature in test.features:
+            data_for_analysis = feature.apply_to_data(data_for_analysis)
+
+        if not self._is_valid_for_analysis(data_for_analysis, test):
+            logger.warning("Data is not valid for the analysis!")
+            return test_result
+
+        # get control and treatment values for the kpi
+        control          = test.variants.get_variant(data_for_analysis, test.variants.control_name)[test.kpi.name]
```
gbordyugov, line 104: those lines look much cleaner now, great job!

gbordyugov (commented):
just a couple of little comments

expan/core/experiment.py
shansfolder: yes it is used to filter out duplicates. Should we still do the same here?

expan/core/experiment.py
shansfolder: Not sure. I think so...I can take a look.

expan/core/experiment.py
shansfolder: good point. I will fix it.

expan/core/experiment.py
shansfolder: good point. I will fix it.

tests/tests_core/test_statistical_test.py
shansfolder: good point. Fixed.

expan/core/experiment.py
gbordyugov: I think that it's quite often the case that data has duplicates — I think we shouldn't break things that we don't have to break and keep at least a resemblance of backward compatibility in terms of functionality.

shansfolder: oh I think I misunderstood.

Here ```entity``` should not have duplicates because we do our analysis based on this entity level. If there is duplicates, something is wrong when we group the data. And I think we should raise an error if this happens.

expan/core/experiment.py
shansfolder: With the amount of data we have, I feel the runtime is similar. I can't feel the difference.

But I think ```np.unique``` is better for our ```pd.Series``` because it works on np array level.

expan/core/early_stopping.py
```diff
@@ -1,73 +1,76 @@
...
+    :type  spending_function: str
+    :param estimated_sample_size: sample size to be achieved towards the end of experiment
+    :type  estimated_sample_size: int
+    :param alpha: type-I error rate
+    :type  alpha: float
+    :param cap: upper bound of the adapted z-score
+    :type  cap: int
+
+    :return: results of type EarlyStoppingTestStatistics
+    :rtype:  EarlyStoppingTestStatistics
```
gbordyugov, line 61: nice documentation, great job!

expan/core/early_stopping.py
```diff
@@ -1,73 +1,76 @@
...
     """
     # Checking if data was provided
     if x is None or y is None:
```
gbordyugov: This check doesn't seem to be needed, since the lines below ensure that `x` and `y` are instances of `Series`, `array` or `list` classes.

expan/core/early_stopping.py
```diff
@@ -106,100 +109,75 @@ def group_sequential(x,
...
-        sm = pickle.load(open(compiled_model_file, 'rb'))
-    else:
-        sm = StanModel(file=model_file)
-        with open(compiled_model_file, 'wb') as f:
-            pickle.dump(sm, f)
-    return sm
-
-cache_sampling_results = False
-sampling_results = {} # memoized sampling results
+    return [HDImin, HDImax]
```
gbordyugov: Why a list instead of a tuple?

expan/core/early_stopping.py
```diff
@@ -106,100 +109,75 @@ def group_sequential(x,
...
+    :param num_iters: number of iterations of sampling
+    :type  num_iters: int
+    :param inference: 'sampling' for MCMC sampling method or 'variational' for variational inference
+    :type  inference: str
+
+    :return: the posterior samples, sample size of x, sample size of y, absolute mean of x, absolute mean of y
+    :rtype:  tuple[array-like, array-like, array-like, float, float]
     """
     # Checking if data was provided
     if x is None or y is None:
```
gbordyugov: The same here, the later type checks seem to make this check against `None` obsolete.

expan/core/early_stopping.py
```diff
@@ -316,40 +292,107 @@ def f(x, y):
...
+    """
+    # check of traces type, if traces is not dict "traces['alpha']" will return KeyError
+    if not isinstance(traces, dict):
+        raise TypeError("Traces statistics is not a dictionary and does not contain alpha or delta statistics.")
+
+    if distribution == 'normal':
+        return traces['alpha']
+    elif distribution == 'poisson':
+        variance = np.nanmean(np.array(traces['delta']))
+        return traces['delta'] / np.sqrt(np.absolute(variance))
```
gbordyugov, line 372: Just a remark for the future work (not part of this pull request):

1. for a normal distribution, the returned effect size is already normalized
2. for a Poisson distribution, we have to normalize it by hand

Maybe we should stick to one particular scheme and return a non-normalized effect size from the normal model (or the normalized from the Poisson model).

expan/core/experiment.py
```diff
@@ -1,280 +1,222 @@
...
+        :type  testmethod: str
+        :param **worker_args: additional arguments for the analysis method
+        :return: statistical result of the test
+        :rtype: StatisticalTestResult
+        """
+        if not isinstance(test, StatisticalTest):
+            raise TypeError("Statistical test should be of type StatisticalTest.")
+
+        if 'entity' not in self.data.columns:
+            raise RuntimeError("There is no 'entity' column in the data.")
```
gbordyugov: A `KeyError` is usually raised when you try to do smth like `df['non_existing_column_name']` and an `AttributeError` if you're doing `df.non_existing_column_name`.

expan/core/experiment.py
```diff
@@ -1,280 +1,222 @@
...
+        if self.data.entity.duplicated().any():
+            raise ValueError('Entities in data should be unique.')
+
+        if test.variants.variant_column_name not in self.data.columns:
+            raise RuntimeError("There is no '{}' column in the data.".format(test.variants.variant_column_name))
+        if test.variants.treatment_name not in np.unique(self.data[test.variants.variant_column_name]):
+            raise RuntimeError("There is no treatment with the name '{}' in the data.".format(test.variants.treatment_name))
+        if test.variants.control_name not in np.unique(self.data[test.variants.variant_column_name]):
+            raise RuntimeError("There is no control with the name '{}' in the data.".format(test.variants.control_name))
+
```
gbordyugov: This really looks very meticulous to me, thanks a lot for this level of attention to the details!

expan/core/early_stopping.py
daryadedik: good question! I think we don't need to change to a list here, since we can do zip on list and tuple and it will return the same result.

expan/core/early_stopping.py
daryadedik: Agree! Will remove that check.

expan/core/early_stopping.py
daryadedik: Agree! Will remove that check.

expan/core/experiment.py
```diff
@@ -1,280 +1,222 @@
...
+        """ Check whether the quality of data is good enough to perform analysis. Invalid cases can be:
+        1. there is no data
+        2. the data does not contain all the variants to perform analysis
+        
+        :type data: DataFrame
+        :type test: StatisticalTest
+        :rtype: bool 
         """
-        if df is None:
+        if data is None:
```
gbordyugov: I don't really see a point in checking against `None`. Do we have a good reason to expect a `None` there? Why don't we check  for `data=5` or `data='a string'` then?

expan/core/results.py
```diff
@@ -0,0 +1,117 @@
...
+    :param ci: a dict where keys are percentiles and values are the corresponding value for the statistic.
+    :type  ci: dict
+    """
+    def __init__(self, control_statistics, treatment_statistics, delta, ci, p, statistical_power):
+        super(SimpleTestStatistics, self).__init__(control_statistics, treatment_statistics)
+        self.delta               = delta
+        self.p                   = p
+        self.statistical_power   = statistical_power
+        # TODO: think of structure {p: v} (the same as ci)
+        self.confidence_interval = [{'percentile': p, 'value': v} for (p, v) in ci.items()]
```
gbordyugov: Didn't we have a class for `ConfidenceInterval`?

expan/core/results.py
```diff
@@ -0,0 +1,117 @@
...
+    :type  statistical_test_results: list[StatisticalTestResult]
+    :param correction_method: method used for multiple testing correction. Possible values are:
+                              "none": no correction
+                              "bh": benjamini hochberg correction
+                              "bf": bonferroni correction
+    :type  correction_method: str
+    """
+    def __init__(self, statistical_test_results, correction_method="none"):
+        self.statistical_test_results = statistical_test_results
+        if correction_method not in ["none", "bh", "bf"]:
```
gbordyugov: Why not stick to something like an `Enum`?

expan/core/statistics.py
```diff
@@ -1,83 +1,78 @@
...
-    Returns:
-        DeltaStatistics object
+    :type: relative: boolean
+    
+    :return: results of type SimpleTestStatistics
+    :rtype: SimpleTestStatistics
     """
-    # Checking if data was provided
+    # Check if data was provided and it has correct format
     if x is None or y is None:
```
gbordyugov, line 50: Those checks are made obsolete by the later ones

expan/core/results.py
daryadedik: So, for analysis service it's ok to have parameters for correction as 0, 1 or 2. We wanted to leave at least a short names for correction for somebody who can use expan from outside Zalando. I also don't think we need to have a class here. Or you have some advantages in mind for doing it as Enum?

expan/core/statistics.py
daryadedik: Sure!

expan/core/util.py
```diff
@@ -1,172 +1,117 @@
...
-
-  >>> scale_range([1,3,5], new_min=0.5)
-  array([ 0.5 ,  0.75,  1.  ])
-
-  >>> scale_range([1,3,5], old_min=1, old_max=4)
-  array([ 0.        ,  0.66666667,  1.        ])
-
-  >>> scale_range([5], old_max=4)
-  array([ 1.])
+    return [item[lookup_key] for item in items if item[condition_key] == condition_value][0]
```
gbordyugov: this one is prone to throwing `KeyError`s

expan/core/results.py
gbordyugov: Each time you add a new correction type or change an existing one, you'll have to find all occurences of, say, `'bh'` in the code. A dedicated `Enum`-type class would be the unique source of truth.

expan/core/experiment.py
daryadedik: Good point. We will change that!

expan/core/experiment.py
shansfolder: Because this method is used for subgroup data. After applying ```FeatureFilter``` to the overal data, the data frame could be ```None```. It could be worth checking at least.

expan/core/results.py
shansfolder: We had a class previously, but then we found that it leads to two times of the fields:
```
"confidence_interval": {
    "confidence_interval" : [
        {"percentile": 2.5, "value": -1},
        {"percentile": 97.5.5, "value": 1}
    ]
}

Comment (shansfolder) [reply expan/core/results.py]: The ```TODO``` here is to think of whether we want:
```
    "confidence_interval" : [
        {"percentile": 2.5, "value": -1},
        {"percentile": 97.5.5, "value": 1}
    ]
```
or
```
    "confidence_interval" : [
        {"2.5": -1},
        {"97.5", 1}
    ]
```

File: expan/core/results.py (hunk 16 above)
Comment (shansfolder) [reply expan/core/results.py]: @gbordyugov yes I agree. One more question: do you also think we need to create a Enum for the delta methods? We currently maintain a lookup table for "group_sequential", "bayes_factor", "fixed_horizon", etc

File: expan/core/util.py (hunk 18 above)
Comment (shansfolder) [reply expan/core/util.py]: Yes. we have created new units tests in ```test_util.py``` which covers this case.

File: expan/core/early_stopping.py (hunk 11 above)
Comment (shansfolder) [reply expan/core/early_stopping.py:372]: good point! I've created a issue #200 to keep track of it. :)

File: expan/core/early_stopping.py (hunk 10 above)
Comment (shansfolder) [reply expan/core/early_stopping.py]: @gbordyugov yes technically the check for ```None``` is redundant, but I would argue that we need this repetition to have clearer error message for easier debugging.

If its ```None```, we know the error is no data provided.
If its wrongly typed, we have different message indicating it.

Comment (shansfolder) [reply expan/core/early_stopping.py]: Also, the additional check for None is nothing in terms of runtime.

File: expan/core/statistics.py (hunk 17 above)
Comment (shansfolder) [reply expan/core/statistics.py:50]: That's true... Well, still, I believe there is more advantages to keep it.

It might covers some tricky usage (e.g. open source user uses ```statistics.delta()``` directly). And the runtime increase to remove it is almost nothing.

File: expan/core/experiment.py (hunk 14 above)
Comment (gbordyugov) [reply expan/core/experiment.py]:
```
In [25]: df
Out[25]:
   a
0  1
1  2
2  3

In [26]: df[df.a == 4]
Out[26]:
Empty DataFrame
Columns: [a]
Index: []

In [27]: df[df.a == 4] is None
Out[27]: False
```

expan/core/results.py
gbordyugov: I see, thank you for clarifying this one for me.

expan/core/results.py
gbordyugov: Yes, that would be another application of this strategy, totally agree!

expan/core/experiment.py
shansfolder: I see. Thanks. Will remove it.

expan/core/results.py
daryadedik: the problem I see here with structure {'2.5': 0.5, '97.5': 0.4} is that everywhere where we need to access those fields we always need to know those numbers (2.5 and 97.5). If they are fixed, then it's ok-ish, but things can change if we decide to have another percentiles (for example) or add some intermediate (we allowed that before as far as I remember).

expan/core/experiment.py
shansfolder: @gbordyugov @daryadedik Actually we **have to** use ```pd.unique()``` instead of ```set()``` on pandas DataFrame.

I just run the method on test data with 8M rows. 
```set()``` hangs forever without any error message (it's also the bug Ievgen discovered in Expan Service).
```np.unique()``` takes 10+ seconds.
```pd.unique()``` takes 0.3 seconds.

expan/core/results.py
shansfolder: you are right. Then I think we can stay with ``` {"percentile": 2.5, "value": -1}```.

shansfolder:
@gbordyugov you can still review it in the history of merge pull request, but I am afraid its too big to review anyways. Maybe checkout the code directly is easier.

I am gonna merge it now, without releasing. So we can still discuss when you're back.

PR #195, opened 2018-02-21
Make result classes JSON serializable

shansfolder:


expan/core/util.py
```diff
@@ -1,10 +1,20 @@
...
 from warnings import warn
 
 import numpy as np
 import pandas as pd
 from datetime import datetime, timedelta
 
 
+class JsonSerializable(object):
+    def toJson(self):
+        return json.dumps(self, default=lambda o: o.name if isinstance(o, Enum) else o.__dict__, sort_keys=True, indent=4)
```
gbordyugov, line 12: Would this work too for nested `JsonSerializable`, i.e. if one instance of `JsonSerializable` has another instance of `JsonSerializable` as one of its fields?

tests/tests_core/test_json_serializer.py
```diff
@@ -0,0 +1,86 @@
...
+                                                         corrected_p,
+                                                         statistical_power,
+                                                         corrected_stop)
+
+
+        test_results = [CorrectedTestStatistics(simple_stats, simple_stats_corrected),
+                        CorrectedTestStatistics(es_stats, es_stats_corrected)]
+        statistical_test_results = MultipleTestSuiteResult(test_results, MultipleTestingCorrectionMethod.no_correction)
+
+        js_result = statistical_test_results.toJson()  # no error/exception should be raise
```
gbordyugov: This certainly doesn't throw an exception/error, but do all fields of `statistical_test_results` get properly serialized in a recursive manner?

shansfolder: yes it works recursively because of the following:

```
class JsonSerializable(object):
    def toJson(self):
        return json.dumps(self, default=lambda o: o.name if isinstance(o, Enum) else o.__dict__, sort_keys=True, indent=4)
```

It actually can only serialize at a flat level and throw an error saying it can't be serialzed without the above lambda.

expan/core/util.py
shansfolder: yes it works. you can run ```pytest -s test_json_serializer.py``` to see the output of those nested result classes.

PR #197, opened 2018-02-27
Adapt statistics.py

shansfolder:


expan/core/statistics.py
```diff
@@ -318,129 +228,50 @@ def bootstrap(x, y, func=_delta_mean, nruns=10000, percentiles=[2.5, 97.5],
...
+    :param n2: size of second sample
+    :type  n2: int
+
+    :return: pooled standard deviation
+    :type: float
     """
     if not (0.5 < (std1 ** 2) / (std2 ** 2) < 2.):
-        warnings.warn('Sample variances differ too much to assume that '
-                      'population variances are equal.')
+        warnings.warn('Sample variances differ too much to assume that population variances are equal.')
```
daryadedik, line 249: Do we need this warning in scalyr? Maybe it make sense to add it to logger?

shansfolder: good point

gbordyugov (approved):
++

PR #210, opened 2018-04-11
raise ValueError on zero pooled std for power calculations

gbordyugov:
raise ValueError on zero pooled std for power calculations


shansfolder (approved):
Very interesting. Thanks! Did you encounter this from real data somewhere?

daryadedik:
👍

PR #213, opened 2018-05-24
Experiment data restructure

daryadedik:
I restructured data available for the analysis. Instead of data per experiment, we provided data per statistical test. This is done because we want to do all necessary business related grouping (like kpi per session or per entity) in expan service and provide grouped data to each statistical test.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -67,7 +67,6 @@ Running an analysis is very simple:
 
 Currently ``analyze_statistical_test`` supports 4 test methods: ``fixed_horizon`` (default), ``group_sequential``, ``bayes_factor`` and ``bayes_precision``.
 All methods requires different additional parameters.
-``include_data`` is False by default, meaning the result statistical test result will not contain analysis data.
 
 If you would like to change any of the default values, just pass them as parameters to delta. For example:
 
@@ -144,7 +143,6 @@ An example of the result is shown below:
             }
         },
         "test": {
-            "data": null,
             "features": [],
             "kpi": {
                 "name": "revenue"
@@ -223,7 +221,6 @@ Following is an example of the analysis result of statistical test suite:
         "results": [
             {
                 "test": {
-                    "data": null,
                     "features": [
                         {
                             "column_name": "device_type",
@@ -294,7 +291,6 @@ Following is an example of the analysis result of statistical test suite:
             },
             {
                 "test": {
-                    "data": null,
                     "features": [
                         {
                             "column_name": "device_type",
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -67,6 +67,7 @@ Running an analysis is very simple:
 
 Currently ``analyze_statistical_test`` supports 4 test methods: ``fixed_horizon`` (default), ``group_sequential``, ``bayes_factor`` and ``bayes_precision``.
 All methods requires different additional parameters.
+``include_data`` is False by default, meaning the result statistical test result will not contain analysis data.
 
 If you would like to change any of the default values, just pass them as parameters to delta. For example:
 
@@ -76,7 +77,7 @@ If you would like to change any of the default values, just pass them as paramet
 	exp.analyze_statistical_test(test, test_method='group_sequential', estimated_sample_size=1000)
 	exp.analyze_statistical_test(test, test_method='bayes_factor', distribution='normal')
 
-Here is the list of addtional parameters.
+Here is the list of additional parameters.
 You may also find the description in our :ref:`API <modindex>` page.
 
 *fixed_horizon* is the default method:
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -15,7 +15,7 @@ First, let's generate some random data for the tutorial.
 
 ``data`` is a pandas DataFrame.
 It must contain a column for entity identifier named **entity**,
-a column for variant, and one column per kpi/feature.
+a column for variant, and one column per kpi/feature. Each data is constructed per each Statistical Test.
 
 ``metadata`` is a python dict. It should contain the following keys:
 
@@ -36,11 +36,10 @@ To use ExpAn for analysis, you first need to create an ``Experiment`` object.
 .. code-block:: python
 
     from expan.core.experiment import Experiment
-    exp = Experiment(data=data, metadata=metadata)
+    exp = Experiment(metadata=metadata)
 
 This ``Experiment`` object has the following parameters:
 
-	* ``data``: A data you want to run experiment for. An example of the data structure. Described above.
 	* ``metadata``: Specifies an experiment name as the mandatory and data source as the optional fields. Described above.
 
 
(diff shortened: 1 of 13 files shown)


docs/tutorial.rst
```diff
@@ -55,7 +54,7 @@ Each statistical test consist of one kpi, treatment and control variant names, a
 
     kpi = KPI('normal_same')
     variants = Variants(variant_column_name='variant', control_name='B', treatment_name='A')
-    test = StatisticalTest(kpi=kpi, features=[], variants=variants)
+    test = StatisticalTest(data=data, kpi=kpi, features=[], variants=variants)
```
shansfolder, line 57: ahaha you made it the first para in the end!

expan/core/experiment.py
```diff
@@ -56,30 +55,30 @@ def analyze_statistical_test(self, test, test_method='fixed_horizon', **worker_a
         if not isinstance(test, StatisticalTest):
             raise TypeError("Statistical test should be of type StatisticalTest.")
 
-        if 'entity' not in self.data.columns:
+        if 'entity' not in test.data.columns:
             raise KeyError("There is no 'entity' column in the data.")
-        if self.data.entity.duplicated().any():
+        if test.data.entity.duplicated().any():
```
shansfolder: should we move the duplication check after we perform the feature filter?

expan/core/experiment.py
```diff
@@ -37,17 +34,19 @@ def __init__(self, data, metadata):
         }
 
     def __str__(self):
-        return 'Experiment "{:s}" with {:d} entities.'.format(self.metadata['experiment'], len(self.data))
+        return 'Performing "{:s}" experiment.'.format(self.metadata['experiment'])
 
 
-    def analyze_statistical_test(self, test, test_method='fixed_horizon', **worker_args):
+    def analyze_statistical_test(self, test, test_method='fixed_horizon', include_data=False, **worker_args):
```
shansfolder, line 40: I would rename ```include_data``` to ```exclude_data_in_result```, and in the end of this method you can just write:

```if exclude_data_in_result: del test.data```

File: expan/core/results.py [hunk 4]
```diff
@@ -99,7 +99,7 @@ def __init__(self, original_test_statistics, corrected_test_statistics):
 class StatisticalTestResult(JsonSerializable):
     """ This class holds the results of a single statistical test.
     
-    :param test: information about the statistical test
+    :param test: information about the statistical test (without data)
```
shansfolder: I think we don't need to write this, but this is very minor :)

expan/core/version.py
```diff
@@ -1,6 +1,6 @@
 import subprocess
 
-__version__ = "1.0.1"
+__version__ = "1.1.0"
```
shansfolder: ah you don't do it here. let's do it via bumpversion: http://expan.readthedocs.io/en/latest/contributing.html#release

expan/core/experiment.py
```diff
@@ -149,7 +151,7 @@ def analyze_statistical_test_suite(self, test_suite, test_method='fixed_horizon'
         # test_suite_result hold statistical results from all statistical tests
         test_suite_result = MultipleTestSuiteResult([], test_suite.correction_method)
         for test in test_suite.tests:
-            original_analysis = self.analyze_statistical_test(test, test_method, **worker_args)
+            original_analysis = self.analyze_statistical_test(test, test_method, True, **worker_args)
```
shansfolder, line 153: True means you will include the data in result?

expan/core/experiment.py
```diff
@@ -164,20 +166,24 @@ def analyze_statistical_test_suite(self, test_suite, test_method='fixed_horizon'
             new_worker_args['alpha'] = corrected_alpha
 
             for test_index, test_item in enumerate(test_suite_result.results):
-                if test_item.result.original_test_statistics: # result can be None if not enough entities
+                if test_item.result.original_test_statistics:  # result can be None if not enough entities
                     original_analysis = test_suite_result.results[test_index]
-                    corrected_analysis = self.analyze_statistical_test(test_item.test, test_method, **new_worker_args)
+                    corrected_analysis = self.analyze_statistical_test(test_item.test, test_method,
```
shansfolder, line 170: here it is True as well...

expan/core/version.py
daryadedik: @shansfolder ok, got it!

docs/tutorial.rst
daryadedik: @shansfolder yes, I decided to keep the priority of the parameters: first and most important is data, then kpis, then features (we can do testing without them).

expan/core/experiment.py
daryadedik: @shansfolder thanks, will change that!

expan/core/experiment.py
daryadedik: done!

expan/core/results.py
daryadedik: removed!

shansfolder (approved):

docs/tutorial.rst
```diff
@@ -15,7 +15,7 @@ First, let's generate some random data for the tutorial.
 
 ``data`` is a pandas DataFrame.
 It must contain a column for entity identifier named **entity**,
-a column for variant, and one column per kpi/feature.
+a column for variant, and one column per kpi/feature. Each data is constructed per each Statistical Test.
```
gbordyugov, line 18: I find this addition slightly confusing. What is meant by 'each data'?

expan/core/experiment.py
gbordyugov: This logic appears strange to me, to be honest. Why would `analyze_statistical_test` want to touch the inner fields of `test`?

expan/core/experiment.py
```diff
@@ -114,6 +111,10 @@ def analyze_statistical_test(self, test, test_method='fixed_horizon', **worker_a
         # run the test method
         test_statistics = worker(x=treatment_data, y=control_data)
         test_result.result = test_statistics
+
+        # remove data from the result test metadata
+        if not include_data:
```
gbordyugov, line 116: Why?

expan/core/experiment.py
```diff
@@ -274,8 +279,8 @@ def _quantile_filtering(self, kpis, percentile, threshold_type):
         :rtype: pd.Series
         """
         method_table = {'upper': lambda x: x > threshold, 'lower': lambda x: x <= threshold}
-        flags = pd.Series(data=[False]*len(self.data))
-        for column in self.data[kpis].columns:
-            threshold = np.percentile(self.data[column], percentile)
-            flags = flags | self.data[column].apply(method_table[threshold_type])
+        flags = pd.Series(data=[False]*len(data))
+        for column in data[kpis].columns:
+            threshold = np.percentile(data[column], percentile)
```
gbordyugov, line 284: Are you sure that `threshold` in the lambda is properly updated here? Maybe pass it along with `x` to be sure?

docs/tutorial.rst
daryadedik: @gb oh yes, my mistake, will change that.

expan/core/experiment.py
daryadedik: because data can be huge, we just don't want it to be present in the results. 

Here include_data is sort of "hidden" parameter, I need it to only be used by analyze_statistical_test_suite. When we run analyze_statistical_test_suite with include_data=True it means it will not delete data because we need those data when we call analyze_statistical_test again in line 170. We delete data for suites in line 180-181. When we call only analyze_statistical_test it will delete data for 1 test because include_data=False.

PR #214, opened 2018-05-25
Update outlier filter on derived kpis

shansfolder:


daryadedik:
@sh all looks good! thanks!

PR #218, opened 2018-05-30
Removed deep copy of the data in statistical test construction

daryadedik:


shansfolder (approved):

gbordyugov:
what's the reason for the deep copy?

shansfolder:
@gbordyugov since each ```StatisticalTest``` has a ```data``` object of its own, and we are passing the reference of same data frame around (in expan service), the initial idea of the ```deepcopy``` is to make each test own a clean copy of data.

shansfolder:
@gbordyugov but then we run into memory issues because there are lots of tests for subgroups, deepcopy leads to memory error, thus we remove the copy in this PR.

PR #219, opened 2018-05-31
Remove null analysis results from the analysis results files

daryadedik:


expan/core/experiment.py
```diff
@@ -153,12 +153,13 @@ def analyze_statistical_test_suite(self, test_suite, test_method='fixed_horizon'
...
-            combined_result = CombinedTestStatistics(original_analysis.result, original_analysis.result)
-            original_analysis.result = combined_result
-            test_suite_result.results.append(original_analysis)
+            # do not include results of the test where statistical power is -1 and if the results are None
+            if original_analysis.result and original_analysis.result.statistical_power != -1:
+                combined_result = CombinedTestStatistics(original_analysis.result, original_analysis.result)
+                original_analysis.result = combined_result
+                test_suite_result.results.append(original_analysis)
+            else:
+                logger.warning("Analysis results are excluded from the result file because they contain Null values.")
```
shansfolder, line 162: there can be two possibilities here. 1. as what you wrote. 2. ``` original_analysis.result is None``` (because of only 1 data in the subgroup)

daryadedik: is it ok if I write then smth like that "Analysis results are excluded from the result file because 1: analysis data is not valid for the analysis, or 2: analysis results contain Null values" or should be more detailed?

shansfolder: ah sorry I think what you wrote here is fine

shansfolder (approved):

PR #221, opened 2018-06-12
Fix for the possibility that both variances are zero

aaron-mcdaid-zalando:
If both variances are exactly zero, we shouldn't complain about them being too different. This PR fixes that. This will be relevant in some of our small numerical tests in future, and of course it could be relevant for some very small real world datasets

### DIFF
--- a/expan/core/statistics.py
+++ b/expan/core/statistics.py
@@ -240,7 +240,8 @@ def pooled_std(std1, n1, std2, n2):
     :return: pooled standard deviation
     :type: float
     """
-    if not (0.5 < (std1 ** 2) / (std2 ** 2) < 2.):
+    if (std1 ** 2) >   2.0*(std2 ** 2) or \
+       (std1 ** 2) <   0.5*(std2 ** 2):
         warnings.warn('Sample variances differ too much to assume that population variances are equal.')
         logger.warning('Sample variances differ too much to assume that population variances are equal.')
 


expan/core/statistics.py
```diff
@@ -240,7 +240,7 @@ def pooled_std(std1, n1, std2, n2):
     :return: pooled standard deviation
     :type: float
     """
-    if not (0.5 < (std1 ** 2) / (std2 ** 2) < 2.):
+    if std1 != std2 and not (0.5 < (std1 ** 2) / (std2 ** 2) < 2.):
```
gbordyugov: if they're both zero, this would throw a division by zero error, wouldn't it?

aaron-mcdaid-zalando: It wouldn't. The whole point of this PR is to avoid exactly that

If they are both zero, the first part of the and will fail and therefore the `if` isn't used and therefore there is no warning.

The general idea here is that if the two stds are equal to each other (regardless of whether it's zero or not), then we don't need to trigger the warning that is inside the if

gbordyugov: Oh, sorry, I misread the 'not-equal' part. What then if `std2 == 0` is true?

aaron-mcdaid-zalando (commented):
@gbordyugov 
(sorry for losing your last comment about division by zero. I misused github a little)

Anyway, this version doesn't use division at all

shansfolder:
@aaron-mcdaid-zalando: I think @gbordyugov meant this line: ```std1 != std2 and not (0.5 < (std1 ** 2) / (std2 ** 2) < 2.)```. When ```std2==0``` and ```std1!=0```, it will raise a division by zero error. 

Maybe we should add a tiny jitter to 0 variance (so it becomes 0.000001)? and log the warning?

shansfolder:
this is not coming from your code though. :)

aaron-mcdaid-zalando:
@shansfolder , a few minutes ago a pushed a new version here that doesn't use division at all

This whole thing seems strange to me though. Do you want to test for this? And if so, perhaps there is a better way to do it

shansfolder:
@aaron-mcdaid-zalando ah yeah now I see the new version. I like it. :)

shansfolder (approved):

gbordyugov (approved):

aaron-mcdaid-zalando:
@gbordyugov , @shansfolder 

I'm "not authorized to merge this pull request."

Should I, `aaron-mcdaid-zalando`, be added to a list?

shansfolder:
@aaron-mcdaid-zalando ```aaron-mcdaid-zalando``` invited. you should receive an email.

PR #222, opened 2018-06-12
More powerful derived kpis

aaron-mcdaid-zalando:
Previously, a derived KPI specified as  `X/Y` would be treated by dividing each element of `X` by the *average* value of `Y`, and then analysis would proceed on these adjusted values:

     adjusted_i = x_i / mean(Y)

This, correctly, meant that the `mean(Adjusted)` was equal the weighted mean of all the `x_i/y_i`. However the variance was overestimated and therefore less powerful. This PR improves the power

The priority in this PR is `make_delta`. This PR doesn't not attempt to apply this fix to `make_group_sequential`, as that will soon be removed and replaced by other changes that are coming. Bootstrapping is also not affected by this PR (but perhaps it should be!)

### DIFF
--- a/expan/core/early_stopping.py
+++ b/expan/core/early_stopping.py
@@ -41,8 +41,8 @@ def make_group_sequential(spending_function='obrien_fleming', estimated_sample_s
     def go(x, y, x_denominators=1, y_denominators=1):
 
         # these next too lines are wrong, but they are bug-compatible with v0.6.13 !
-        x = x / np.mean(x_denominators)
-        y = y / np.mean(y_denominators)
+        x = x / np.nanmean(x_denominators)
+        y = y / np.nanmean(y_denominators)
 
         return group_sequential(x, y, spending_function, estimated_sample_size, alpha, cap)
     return go
@@ -242,8 +242,8 @@ def _bayes_sampling(x, y, distribution='normal', num_iters=25000, inference="sam
 def make_bayes_factor(distribution='normal', num_iters=25000, inference='sampling'):
     """ Closure method for the bayes_factor"""
     def f(x, y, x_denominators = 1, y_denominators = 1):
-        x = x / np.mean(x_denominators)
-        y = y / np.mean(x_denominators)
+        x = x / np.nanmean(x_denominators)
+        y = y / np.nanmean(y_denominators)
         return bayes_factor(x, y, distribution, num_iters, inference)
     return f
 
@@ -303,8 +303,8 @@ def bayes_factor(x, y, distribution='normal', num_iters=25000, inference='sampli
 def make_bayes_precision(distribution='normal', posterior_width=0.08, num_iters=25000, inference='sampling'):
     """ Closure method for the bayes_precision"""
     def f(x, y, x_denominators = 1, y_denominators = 1):
-        x = x / np.mean(x_denominators)
-        y = y / np.mean(x_denominators)
+        x = x / np.nanmean(x_denominators)
+        y = y / np.nanmean(y_denominators)
         return bayes_precision(x, y, distribution, posterior_width, num_iters, inference)
     return f
 
--- a/expan/core/early_stopping.py
+++ b/expan/core/early_stopping.py
@@ -38,7 +38,14 @@ def obrien_fleming(information_fraction, alpha=0.05):
 
 def make_group_sequential(spending_function='obrien_fleming', estimated_sample_size=None, alpha=0.05, cap=8):
     """ A closure to the group_sequential function. """
-    return lambda x, y: group_sequential(x, y, spending_function, estimated_sample_size, alpha, cap)
+    def go(x, y, x_denominators=1, y_denominators=1):
+
+        # these next too lines are wrong, but they are bug-compatible with v0.6.13 !
+        x = x / np.mean(x_denominators)
+        y = y / np.mean(y_denominators)
+
+        return group_sequential(x, y, spending_function, estimated_sample_size, alpha, cap)
+    return go
 
 
 def group_sequential(x, y, spending_function='obrien_fleming', estimated_sample_size=None, alpha=0.05, cap=8):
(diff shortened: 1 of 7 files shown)


aaron-mcdaid-zalando:
Don't merge yet, even though the tests pass! I found some bugs that weren't caught by any test. As well as fixing them, I think I'll add some tests. Basically, the closures/lambdas return by the four "worker factories" aren't really tested

aaron-mcdaid-zalando:
OK, I'm happier now with this version of the code. Let's start reviewing!

I'm going to write another test though, to quickly run a large number of A/A tests (with small sample size) to confirm that the p-values are nicely distributed

expan/core/statistics.py
```diff
@@ -85,16 +110,28 @@ def delta(x, y, assume_normal=True, alpha=0.05, min_observations=20, nruns=10000
         if assume_normal:
             logger.info("The distribution of two samples is assumed normal. "
                         "Performing the sample difference distribution calculation.")
-            c_i = normal_sample_difference(x=_x, y=_y, percentiles=percentiles, relative=relative)
+            lots_of_info = normal_sample_weighted_difference(x_numerators=_x, y_numerators=_y, x_denominators=_x_denominators, y_denominators = _y_denominators, percentiles=percentiles, relative=relative)
```
daryadedik: this name of the variable is not straightforward to understand. Maybe some more meaningful name?

expan/core/early_stopping.py
```diff
@@ -38,7 +38,14 @@ def obrien_fleming(information_fraction, alpha=0.05):
 
 def make_group_sequential(spending_function='obrien_fleming', estimated_sample_size=None, alpha=0.05, cap=8):
     """ A closure to the group_sequential function. """
-    return lambda x, y: group_sequential(x, y, spending_function, estimated_sample_size, alpha, cap)
+    def go(x, y, x_denominators=1, y_denominators=1):
+
+        # these next too lines are wrong, but they are bug-compatible with v0.6.13 !
+        x = x / np.nanmean(x_denominators)
+        y = y / np.nanmean(y_denominators)
```
daryadedik, line 45: @aaron-mcdaid-zalando question for my understanding: why we need that devisions?

aaron-mcdaid-zalando: The current system, the "weighting trick", simply divides each numerator by the mean of the denominators. That is what I am replicating here.

For `fixed_horizon`, I'm doing the right thing. For the other three (including this one) I'm simply replicating the existing behaviour.

The big picture is that the interface to a worker has been changed:
  - for a direct kpi, pass the two samples as-is
  - for derived kpis, pass four samples, i.e. the numerators and denominators for control and also for treatment

expan/core/statistics.py
aaron-mcdaid-zalando: Good point :-)  It's quite similar to the delta_statistics. I might just call it `partial_delta_statistics`. I'll push something after my next meeting

expan/core/experiment.py
```diff
@@ -284,6 +286,12 @@ def _get_weights(self, data, test, variant_name):
         number_of_non_zeros_and_nans = len(x) - number_of_zeros_and_nans
         return number_of_non_zeros_and_nans / np.nansum(x) * x
 
+    def _get_denominators(self, data, test, variant_name):
+        if type(test.kpi) is not DerivedKPI:
```
daryadedik: @aaron-mcdaid-zalando I noticed you left _get_weights method, do we need it if we have this one? Also would be nice to have dosctring here.

aaron-mcdaid-zalando (commented):
This change (adding `ddof=1` to some calls) makes a small numerical change, but it's needed to ensure our pooled_variance estimate is correct. This change basically ensures the variance estimate is divided by (n-1), not (n) as is the default in numpy.

See the defininition of `s_i^2` in https://en.wikipedia.org/wiki/Pooled_variance#Definition_and_computation

As a result, I had to make small changes to some tests in this commit also

shansfolder:
@aaron-mcdaid-zalando thanks for this great fix. I'll take a closer look later today, but I have three general comments atm:

1. I merged your other PR, and now this PR has merge conflict with ```master``` branch you might want to resolve.
2. I've given admin to ```aaron-mcdaid-zalando```. You should be able to merge  now.
3. Please also update corresponding documentation in the ```/docs``` folder when this PR is finished.

Thanks a lot again!

aaron-mcdaid-zalando:
I updated documents in a two commits this afternoon

aaron-mcdaid-zalando:
This morning, Monday, I pushed another few important commits to this. If a derived KPI is `5/0` then we should remove it. This might seem crazy, but we sometimes see that in real data. My first commits in this PR were a bit inconsistent about how they handled zeros and NaNs

Simply put: Any datapoint with NaN in the numerator should be discarded. Also, any datapoint with zero or NaN in the denominator should be discarded.

aaron-mcdaid-zalando:
I merged with master (v1.2.5) this morning, and then discovered a bug. So I pushed a fix.

Earlier in this PR, I rewrote `delta` to handle the derived KPIs in a more correct way. But, while doing that, I broke support for pandas Series in the function. Now, I've changed this function so that all inputs are changed to `np.array(dtype=float)` as soon as possible

Tests passing again now

gbordyugov:
My only wish would be to have a little intro to the documentation section explaining the problem this PR is solving.

aaron-mcdaid-zalando:
A few changes pushed since the last review:

- A little extra commentary in the Glossary
- delete some variables that are computed, but no longer used. And delete the `_get_weights` function as it's no longer used
- Improve the testing a little, in particular adding a test where there is a non-zero effect. This is done to confirm that the pvalues are pulled down

expan/core/statistics.py
```diff
@@ -16,10 +18,10 @@ def _delta_mean(x, y):
 
 def make_delta(assume_normal=True, alpha=0.05, min_observations=20, nruns=10000, relative=False):
     """ A closure to the delta function. """
-    return lambda x, y: delta(x, y, assume_normal, alpha, min_observations, nruns, relative)
+    return lambda x, y, x_denominators=1, y_denominators=1: delta(x, y, x_denominators, y_denominators, assume_normal, alpha, min_observations, nruns, relative)
```
gbordyugov: I think that writing out this lambda with `def` would make the code a bit more readable

gbordyugov:
I think we can merge this

expan/core/statistics.py
aaron-mcdaid-zalando: Done. A local `def` named `go`

PR #223, opened 2018-06-21
Performance improvement of variant comparison

shansfolder:


expan/core/experiment.py
```diff
@@ -253,10 +249,11 @@ def _is_valid_for_analysis(self, data, test):
         :return True if data is valid for analysis and False if not
         :rtype: bool 
         """
-        if len(data[data[test.variants.variant_column_name] == test.variants.control_name].dropna()) <= 1:
+        column_name = test.variants.variant_column_name
+        if len(data.loc[data[column_name] == test.variants.control_name, column_name].dropna()) <= 1:
             logger.warning("Control group only contains 1 or 0 entities.")
             return False
-        if len(data[data[test.variants.variant_column_name] == test.variants.treatment_name].dropna()) <= 1:
+        if len(data.loc[data[column_name] == test.variants.treatment_name, column_name].dropna()) <= 1:
```
igusher: do you still need `dropna()`? I didn't try it myself, but I doubt that condition returns True for NAs.
`data[column_name] == test.variants.treatment_name`

shansfolder: no we don't need it:)

shansfolder:
@igusher @gbordyugov @aaron-mcdaid-zalando  It looks good to me. Please confirm to merge and deploy this version.

igusher:
:+1:

aaron-mcdaid-zalando (commented):
Don't do this! This will drop any row with an NA in any column. I think this means any row with zero sessions, as 'revenue-per-session' will be NA (I think) in that case

shansfolder:
@aaron-mcdaid-zalando I've made a commit to handle nans now.

aaron-mcdaid-zalando:
I see the current state includes lots of other changes for `min_observations` and so on. I'm not happy with putting multiple features into one PR, especially given our current problems with quality/testing/correctness

This should be a one-line change

shansfolder:
solved in #224

PR #224, opened 2018-06-22
Counting bugfix and save memory

aaron-mcdaid-zalando:
Save memory by counting only in the 'variant' Series directly.

Also, fix a bug where rows were ignored (for the purpose of this check) if they had NA is any other column

And stop throwing `KeyError` if control/treatment are underrepresented, this was an important issue

And finally, give a `warning` - instead of an `error` - if there are too few datapoints.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -59,10 +59,6 @@ class Experiment(object):
             raise KeyError("There is no 'entity' column in the data.")
         if test.variants.variant_column_name not in test.data.columns:
             raise KeyError("There is no '{}' column in the data.".format(test.variants.variant_column_name))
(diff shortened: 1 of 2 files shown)


shansfolder:
thank you Aaron, please 

1. remove the assertions in line 62-line 65, we don't need it anymore and it will be caught in ```_is_valid_for_analysis``` later.
2. please make ```logger.error("Zero pooled std in compute_statistical_power.")``` to warning level. This is a typo I wrote a while ago.

aaron-mcdaid-zalando:
Thanks for the feedback, Shan. Good catches. I'll push those changes here in a few seconds

shansfolder (approved):

PR #225, opened 2018-06-28
Ensure that outlier detection works if there is NaN in the data

aaron-mcdaid-zalando:
In this PR, NaN will be replaced by the most negative floating point number prior to computing the percentile.


daryadedik (approved):

gbordyugov:
> In this PR, NaN will be replaced by the most negative floating point number prior to computing the percentile.

Why exactly this solution and not smth else, for instance, dropping them altogether?

PR #226, opened 2018-07-02
contributing.rst: Improve the release procedure to ensure that the up…

aaron-mcdaid-zalando:
contributing.rst: Improve the release procedure to ensure that the updated changelog is also included in a release

Previously, the tagged release version never had the correct changelog. For example, this is release 1.2.3, but 1.2.3 is missing from the changelog, and the change info for 1.2.2 is named "Unreleased". https://github.com/zalando/expan/blob/0a87caa73fb591c66ac69e79a17a6963d136a06c/CHANGELOG.md

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -43,12 +43,16 @@ To make a release and **deploy to PyPI**, please follow these steps (we highly s
 
 The flow would then look like follows:
 
-1. :code:`bumpversion (patch|minor)`
-2. :code:`make docs`
-3. :code:`git add CHANGELOG.*`
-4. :code:`git commit -m "update changelog"`
-5. :code:`git push`
-6. :code:`git push --tags`
+1. :code:`bumpversion patch` or :code:`bumpversion minor`
+2. :code:`git describe --tags`, and note this tag name. We will need to edit this tag later.
+3. :code:`make docs`, which will extend the changelog by reading information from github.com/zalando/expan .
+4. :code:`git add CHANGELOG.*`
+5. :code:`git commit --amend --no-edit`
+6. :code:`git show`. Carefully review this commit before proceeding. Ensure the changelog is updated with the expected text, in particular a fully up-to-date version number.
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -43,12 +43,16 @@ To make a release and **deploy to PyPI**, please follow these steps (we highly s
...
 The flow would then look like follows:
 
-1. :code:`bumpversion (patch|minor)`
-2. :code:`make docs`
-3. :code:`git add CHANGELOG.*`
-4. :code:`git commit -m "update changelog"`
-5. :code:`git push`
-6. :code:`git push --tags`
+1. :code:`bumpversion patch` or :code:`bumpversion minor`
+2. :code:`git describe --tags`, and note this tag name. We will need to edit this tag later.
```
gbordyugov, line 47: would you here expect a tag with the version you're trying to release?

docs/contributing.rst
```diff
@@ -43,12 +43,16 @@ To make a release and **deploy to PyPI**, please follow these steps (we highly s
...
+3. :code:`make docs`, which will extend the changelog by reading information from github.com/zalando/expan .
+4. :code:`git add CHANGELOG.*`
+5. :code:`git commit --amend --no-edit`
+6. :code:`git show`. Carefully review this commit before proceeding. Ensure the changelog is updated with the expected text, in particular a fully up-to-date version number.
```
gbordyugov, line 51: so if I'm trying to release 1.5.3, I would expect `1.5.3` in changelog?

docs/contributing.rst
aaron-mcdaid-zalando: Yes

docs/contributing.rst
aaron-mcdaid-zalando: Yes, this works because (as part of the 1.3.1 release) I tweaked the Makefile in order that it uses the correct name of the intended release. See the options `--future-release` in: https://github.com/zalando/expan/blob/852ce9aad809c53316c1fbbbb54d7022b248b24f/Makefile

daryadedik (approved):
looks good! thanks!

PR #227, opened 2018-07-04
OCTO-2622: chi-square test for check of variant split

daryadedik:
This pull request is not ready for review yet: according use of the test should be implemented.

Chi-square test is checked and very slightly adapted compared to previous implementation.

### DIFF
--- a/expan/core/statistics.py
+++ b/expan/core/statistics.py
@@ -549,7 +549,7 @@ def compute_statistical_power(mean1, std1, n1, mean2, std2, n2, z_1_minus_alpha)
     """
 
     # First, check we have enough data for a t-test:
-    if min(n1,n2) < 1 or max(n1,n2) < 2:
+    if min(n1, n2) < 1 or max(n1, n2) < 2:
         return -1
 
     effect_size = mean1 - mean2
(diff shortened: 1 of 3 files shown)


expan/core/statistics.py
```diff
@@ -614,16 +614,72 @@ def compute_p_value(mean1, std1, n1, mean2, std2, n2):
...
+
+
+def chi_square(x, y, min_counts=5):
+    """ Computes chi-square p-value and chi-square test statistics on samples x and y.
+    
+    :param x: sample of the treatment group
+    :type  x: pd.Series or array-like
+    :param y: sample of the control group
+    :type  x: pd.Series or array-like
+    :param min_counts: minimum number of observed and expected frequencies (should be at least 5)
```
gbordyugov, line 639: where does the number `5` come from?

daryadedik: "This test is invalid when the observed or expected frequencies in each category are too small. A typical rule is that all of the observed and expected frequencies should be at least 5." - https://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html

PR #228, opened 2018-07-06
Chi-squared test for the variant split check

daryadedik:
I have removed the previous chi-square test of homogeneity, since this is not exactly what we need.

A test of homogeneity tests whether two (or more) sets of multinomial counts come from different sets of population proportions. Or it tries to find out whether two different unknown populations have the same distribution as each other.
A goodness of fit test is for testing whether a set of multinomial counts is distributed according to a pre-specified set of population proportions. 
The second is what we need: we try to answer the question whether our data or, more specifically,  our variants splits consistent with the expected proportions we claim.

If you have comments or doubts feel free to approach and discuss.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -2,7 +2,6 @@ import logging
 import warnings
 
 import numpy as np
-import pandas as pd
 import copy
 
 import expan.core.early_stopping as es
@@ -311,29 +310,25 @@ class Experiment(object):
         :rtype:  Boolean, float, float
         """
         if not variant_column:
-            logger.warning("Variant split check was cancelled since input variant column is empty or doesn't exist.")
-            return None
+            raise ValueError("Variant split check was cancelled since input variant column is empty or doesn't exist.")
         if not weights:
-            logger.warning("Variant split check was cancelled since input weights are empty or doesn't exist.")
-            return None
+            raise ValueError("Variant split check was cancelled since input weights are empty or doesn't exist.")
         if len(weights) <= 1 or len(variant_column) <= 1:
-            logger.warning("Variant split check was cancelled since input weights or the number if categories "
-                           "is less than 2.")
-            return None
+            raise ValueError("Variant split check was cancelled since input weights or the number if categories "
+                             "is less than 2.")
 
         # Count number of observations per each variant
-        variant_column = pd.Series(variant_column).fillna(0)
+        variant_column = pd.Series(variant_column).dropna(axis=0)
         variants_counts = variant_column.value_counts()
-        observed_counts = pd.DataFrame(variants_counts, columns=['observed']).fillna(0)
+        observed_counts = pd.DataFrame(variants_counts, columns=['observed'])
         # Ensure at least a frequency of 5 at every location in observed_counts, otherwise drop the category
         # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
-        observed_freqs = observed_counts[observed_counts >= min_counts].dropna(axis=0)
+        observed_freqs = observed_counts[observed_counts >= min_counts]
 
         # If all categories were dropped or if there are less than 2 categories left we can't conduct the test.
         if observed_freqs.empty or len(observed_freqs) < 2:
-            logger.warning("Category frequencies are empty or number of categories is less than 2. "
-                           "Chi-square test is not applicable.")
-            return None
+            raise ValueError("Category frequencies are empty or number of categories is less than 2. "
+                             "Chi-square test is not applicable.")
 
         observed_freqs.index.name = 'value'
 
@@ -348,8 +343,7 @@ class Experiment(object):
 
         # If after dropping there are less then two categories we can't conduct the test.
         if len(all_freqs) <= 1:
-            logger.warning("Provide weights for at least two categories. Chi-square test is not applicable.")
-            return None
+            raise ValueError("Provide weights for at least two categories. Chi-square test is not applicable.")
 
         # Compute chi-square and p-value statistics
         chi_square_val, p_val = statx.chi_square(all_freqs['observed'], all_freqs['weight'])
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -331,7 +331,7 @@ class Experiment(object):
 
         # Calculate expected counts given corresponding weights,
         # weights are filtered out of categories which were dropped before.
-        total_count = observed_freqs.sum().sum()
+        total_count = observed_freqs.sum()
         weights = {k: v for (k, v) in weights.items() if k in observed_freqs.index.values}
         expected_freqs = pd.Series(weights)
         expected_freqs *= total_count
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -301,7 +301,8 @@ class Experiment(object):
         """ Tests the consistency of variant split with the hypothesized distribution.
         
         :param variant_column: variant column from the input data frame
-        :param weights: list of weights/split percentage per each variant
+        :param weights: list of dicts with weights/split percentage per each variant 
+                        ([{'value':<variant>,'weight':<weight>}, ...])
         :param min_counts: minimum number of observed and expected frequencies (should be at least 5), see 
                             http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
         :param alpha: significance level, 0.05 by default
@@ -309,11 +310,15 @@ class Experiment(object):
                  False(if split is not consistent with the given split)
         :rtype:  Boolean, float, float
         """
-        if not variant_column or len(variant_column) == 0:
-            logger.warning("Variant split check was cancelled since input variant column is empty or doesn't exist")
+        if not variant_column:
+            logger.warning("Variant split check was cancelled since input variant column is empty or doesn't exist.")
             return None
-        if len(weights) <= 1:
-            logger.warning("Please provide weight for more than pne variant. Variant split check is unsuccessful.")
+        if not weights:
+            logger.warning("Variant split check was cancelled since input weights are empty or doesn't exist.")
+            return None
+        if len(weights) <= 1 or len(variant_column) <= 1:
+            logger.warning("Variant split check was cancelled since input weights or the number if categories "
+                           "is less than 2.")
             return None
 
         # Count number of observations per each variant
@@ -324,9 +329,10 @@ class Experiment(object):
         # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
         observed_freqs = observed_counts[observed_counts >= min_counts].dropna(axis=0)
 
-        # If all categories were dropped we can't conduct the test.
-        if observed_freqs.empty:
-            logger.warning("All variants have frequencies less than 5. Chi-square test is not applicable.")
+        # If all categories were dropped or if there are less than 2 categories left we can't conduct the test.
+        if observed_freqs.empty or len(observed_freqs) < 2:
+            logger.warning("Category frequencies are empty or number of categories is less than 2. "
+                           "Chi-square test is not applicable.")
             return None
 
         observed_freqs.index.name = 'value'
@@ -335,12 +341,17 @@ class Experiment(object):
         total_count = observed_freqs.sum().sum()
         expected_freqs = pd.DataFrame(weights)
 
-        # Join all values in one data frame, so that each observed category counts correspond to the correct expected
+        # Join all values in one data frame, so that each observed category count corresponds to the correct expected
         # count defined by the corresponding weight or split percentage.
-        all_freqs = expected_freqs.join(observed_freqs, on='value')
+        all_freqs = expected_freqs.join(observed_freqs, on='value').dropna(axis=0)
         all_freqs.loc[:, 'weight'] *= total_count
 
+        # If after dropping there are less then two categories we can't conduct the test.
+        if len(all_freqs) <= 1:
+            logger.warning("Provide weights for at least two categories. Chi-square test is not applicable.")
+            return None
+
         # Compute chi-square and p-value statistics
-        chi_square_val, p_val = statx.compute_chi_square(all_freqs['observed'], all_freqs['weight'])
+        chi_square_val, p_val = statx.chi_square(all_freqs['observed'], all_freqs['weight'])
 
         return p_val >= alpha, p_val, chi_square_val
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -323,12 +323,11 @@ class Experiment(object):
         observed_counts = pd.DataFrame(variants_counts, columns=['observed'])
         # Ensure at least a frequency of 5 at every location in observed_counts, otherwise drop the category
         # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
-        observed_freqs = observed_counts[observed_counts >= min_counts]
+        observed_freqs = observed_counts[observed_counts >= min_counts].dropna(axis=0)
 
-        # If all categories were dropped or if there are less than 2 categories left we can't conduct the test.
-        if observed_freqs.empty or len(observed_freqs) < 2:
-            raise ValueError("Category frequencies are empty or number of categories is less than 2. "
-                             "Chi-square test is not applicable.")
+        # If there are less than 2 categories left after dropping counts less than 5 we can't conduct the test.
+        if len(observed_freqs) < 2:
+            raise ValueError("If the number of categories is less than 2 Chi-square test is not applicable.")
 
         # Calculate expected counts given corresponding weights
         total_count = observed_freqs.sum().sum()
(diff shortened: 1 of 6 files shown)


expan/core/experiment.py
```diff
@@ -287,11 +288,70 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        :param min_counts: minimum number of observed and expected frequencies (should be at least 5), see 
+                            http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
+        :param alpha: significance level, 0.05 by default
+        :return: True(if split is consistent with the given split) or
+                 False(if split is not consistent with the given split)
+        :rtype:  Boolean, float, float
+        """
+        if not variant_column:
+            logger.warning("Variant split check was cancelled since input variant column is empty or doesn't exist.")
+            return None
```
igusher: What about raising an exception if calling code doesn't fulfil method's contract? 
It will help to simplify caller code, you won't have to worry about special handling of Nones

expan/core/experiment.py
```diff
@@ -287,11 +288,70 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        if not weights:
+            logger.warning("Variant split check was cancelled since input weights are empty or doesn't exist.")
+            return None
+        if len(weights) <= 1 or len(variant_column) <= 1:
+            logger.warning("Variant split check was cancelled since input weights or the number if categories "
+                           "is less than 2.")
+            return None
+
+        # Count number of observations per each variant
+        variant_column = pd.Series(variant_column).fillna(0)
```
igusher: do you need to fill nas ? won't they be ignored anyway later on?

igusher: or maybe you can drop nas?

expan/core/experiment.py
```diff
@@ -287,11 +288,70 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+
+        # Join all values in one data frame, so that each observed category count corresponds to the correct expected
+        # count defined by the corresponding weight or split percentage.
+        all_freqs = expected_freqs.join(observed_freqs, on='value').dropna(axis=0)
+        all_freqs.loc[:, 'weight'] *= total_count
+
+        # If after dropping there are less then two categories we can't conduct the test.
+        if len(all_freqs) <= 1:
+            logger.warning("Provide weights for at least two categories. Chi-square test is not applicable.")
+            return None
```
igusher: exception, maybe?

expan/core/experiment.py
daryadedik: yes, I don't need it.

expan/core/experiment.py
daryadedik: yes, exceptions probably would be better.

expan/core/experiment.py
```diff
@@ -287,11 +287,65 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
             flags = flags | data[column].apply(method_table[threshold_type])
         return flags
+
+
+    def variant_split_is_consistent_with_expected_distribution(self, variant_column, weights, min_counts=5, alpha=0.05):
+        """ Tests the consistency of variant split with the hypothesized distribution.
+        
+        :param variant_column: variant column from the input data frame
+        :param weights: list of dicts with weights/split percentage per each variant 
+                        ([{'value':<variant>,'weight':<weight>}, ...])
```
gbordyugov: why not `{'control': 0.5, 'treatment': 0.5}` instead of that list of `value-weight` pairs?

daryadedik: because there can be different number of variants and splits ratios (4 variants 25% each, for example).

daryadedik: because there can be different number of treatments (1 control and 2 treatments, for example).

daryadedik:
Note: I am still a bit concerned about the correction we need to perform for multiple testing again (since chi-square is also a statistical test). If we log and peek into the results again, we need to correct alpha again. Maybe we can avoid this by not conducting the statistical test, but rather by introducing the threshold. But I will investigate which threshold will be meaningful.

aaron-mcdaid-zalando:
We could consider an alternative approach entirely. We could simply arrange to store the raw data (the number of assignments) and then do the statistical analysis in some external pipeline connected to our monitoring tool. We might want to do a lot of debugging and experimenting with different analysis methods for this data.

Then, in the longer term, when we understand this issue better and we understand the multiple testing issue, we can consider how much of the analysis can be added into `expan`.

In general, I think we should "gather good data first, ask questions later"

expan/core/experiment.py
gbordyugov: what about `{'control': 0.5, 'treatment_1': 0.25, 'treatment_2': 0.25}`?

daryadedik: I made it easy from expan service point of view: the endpoint returns 

"configuration": {
    "traffic_split_configuration": {
      "variants": [
        {
          "value": "Customization Modal disabled",
          "weight": 1000,
          "is_default": true,
          "image_url": "https://s3.eu-central-1.amazonaws.com/upload-variant-image-bucket/b8fee789-f250-495d-8065-8571e597ac77_bildschirmfoto_2018_04_17_um_18.32.12.png"
        },
        {
          "value": "Customization Modal enabled",
          "weight": 0,
          "is_default": false,
          "image_url": "https://s3.eu-central-1.amazonaws.com/upload-variant-image-bucket/49d35c6d-02c9-4426-9d37-c459fbab12d0_screen_shot_2018_04_17_at_20.16.10.png"
        }
      ],

I just want to grab the necessary json from the output.

igusher: image_url ??
Please focus on lib APIs, make clear boundaries, don't overuse datastructures from the outer world, it may complicate maintenance in a long run.

daryadedik: this was the example of the API return (steering2), I want to use those two fields (value, weight). How else can I get the traffic split?

aaron-mcdaid-zalando: @igusher , @daryadedik , why are you discussing an endpoint? As far as I can see, this code neither consumes nor provides an endpoint. Such an endpoint wouldn't be in this repository anyway

igusher: @aaron-mcdaid-zalando  by API I meant API which Expan lib provides. not some [rest] endpoints from any other  service

daryadedik: @igusher I just misunderstood your comment. I will review my code when I come back.

daryadedik:
In decided to stick with the chi-square test goodness of fit test, because it's very straightforward (you just need a variant column and expected variant weights) and the other: if you test multiple times it's easy to correct with our correction methods: just collect p values and call method from out correction module and get the new alpha to make a decision.

expan/core/experiment.py
```diff
@@ -287,11 +287,64 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        variant_column = pd.Series(variant_column).dropna(axis=0)
+        variants_counts = variant_column.value_counts()
+        observed_counts = pd.DataFrame(variants_counts, columns=['observed'])
+        # Ensure at least a frequency of 5 at every location in observed_counts, otherwise drop the category
+        # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
+        observed_freqs = observed_counts[observed_counts >= min_counts]
+
+        # If all categories were dropped or if there are less than 2 categories left we can't conduct the test.
+        if observed_freqs.empty or len(observed_freqs) < 2:
+            raise ValueError("Category frequencies are empty or number of categories is less than 2. "
```
shansfolder: empty (n=0) is included in n<2.

I would make the error message simpler, or do two assertions if you want the two different messages

expan/core/experiment.py
```diff
@@ -287,11 +287,64 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+
+        # Calculate expected counts given corresponding weights
+        total_count = observed_freqs.sum().sum()
+        expected_freqs = pd.DataFrame.from_dict(weights, orient='index')
+        expected_freqs.columns = ['weight']
+        expected_freqs *= total_count
+
+        # Join all values in one data frame, so that each observed category count corresponds to the correct expected
+        # count defined by the corresponding weight or split percentage.
+        all_freqs = expected_freqs.join(observed_freqs).dropna(axis=0)
```
shansfolder: why do you join ```expected_freqs``` and ```observed_freqs``` into one dataframe ```all_freqs```?

The only use of ```all_freqs``` (I found) is to separate it back into two dataframe in the next lines: ```statx.chi_square(all_freqs['observed'], all_freqs['weight'])```

expan/core/experiment.py
```diff
@@ -287,11 +287,64 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+            raise ValueError("Variant split check was cancelled since input weights or the number if categories "
+                             "is less than 2.")
+
+        # Count number of observations per each variant
+        variant_column = pd.Series(variant_column).dropna(axis=0)
+        variants_counts = variant_column.value_counts()
+        observed_counts = pd.DataFrame(variants_counts, columns=['observed'])
+        # Ensure at least a frequency of 5 at every location in observed_counts, otherwise drop the category
+        # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
+        observed_freqs = observed_counts[observed_counts >= min_counts]
```
shansfolder: If we drop some variant observation here, the length of ```observed_freqs``` will be different than the length of ```weights```. We should catch this mismatch as an exception, or drop the corresponding weight for this variant, instead of silently removing the observations.

shansfolder (changes requested):

expan/core/experiment.py
daryadedik: true, will do it simpler, we don't need to distinguish between two of them, because we can't conduct a test in both cases.

expan/core/experiment.py
daryadedik: I did that because I need the correct corresponding observed and expected count per variant, expected_freqs and observed_freqs are not sorted, so, the values can be mixed. For example, with all_freqs I will have 

   weight  observed
A    25.0        23
B    15.0        18
C    10.0        17
D    40.0        19
E    10.0        23

separately my lists will be 
A    25.0
B    15.0
C    10.0
D    40.0
E    10.0
 and 

  observed
E        23
A        23
D        19
B        18
C        17
 I can sort both as a solution.

daryadedik: sort_index is simpler =)

expan/core/experiment.py
daryadedik: you'r right, just got an error in my unit tests.

daryadedik:
Ready for review!

daryadedik:
@shansfolder @gbordyugov @aaron-mcdaid-zalando please review or approve, I need it for further development.

shansfolder (approved):

shansfolder:
@daryadedik it looks good from my side, but I would suggest to wait for a second reviewer for such new functionality. ;)

daryadedik:
@shansfolder sure.

expan/core/experiment.py
```diff
@@ -287,11 +287,55 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        :param weights: dict with variant names as keys, weights as values
+                        ({<variant_name>:<weight>, ...}
+        :param min_counts: minimum number of observed and expected frequencies (should be at least 5), see 
+                            http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
+        :param alpha: significance level, 0.05 by default
+        :return: True(if split is consistent with the given split) or
+                 False(if split is not consistent with the given split)
+        :rtype:  Boolean, float, float
+        """
+        if not variant_column:
```
aaron-mcdaid-zalando: This should be:

    if not hasattr(variant_column, '__len__'):
          ....

The goal is to ensure that the user has passed in something that has a length, i.e. that `len(x)` works

See https://stackoverflow.com/a/23567686/146041 for more

daryadedik: good point! thanks

aaron-mcdaid-zalando:
Is this true?:

This adds two new function, `chi_square_test_result_and_statistics` and `chi_square`, but those tests are not called anywhere (except in the tests).

If that is true, then this can be safely merged. I'm not in a position to fully review the maths, but I'll give a thumbs up if the above is true

daryadedik:
@aaron-mcdaid-zalando chi_square is called inside of chi_square_test_result_and_statistics. The letter is not called anywhere in expan, except tests

aaron-mcdaid-zalando:
:+1: 

(feel free to make more changes, and ping me again for a thumbs up)

expan/core/experiment.py
```diff
@@ -287,11 +287,55 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+            raise ValueError("Variant split check was cancelled since input weights or the number if categories "
+                             "is less than 2.")
+
+        # Count number of observations per each variant
+        variant_column = pd.Series(variant_column).dropna(axis=0)
+        variants_counts = variant_column.value_counts()
+        observed_counts = pd.DataFrame(variants_counts, columns=['observed'])
+        # Ensure at least a frequency of 5 at every location in observed_counts, otherwise drop the category
+        # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
+        observed_freqs = observed_counts[observed_counts >= min_counts].dropna(axis=0)
```
aaron-mcdaid-zalando: `observed_counts` is a `DataFrame`, not a `Series`, so I don't like `observed_counts >= min_counts`. It could be `observed_counts['observed'] >= min_counts`

Or perhaps `.query` is better, I only recently discovered it in pandas:


    observed_freqs = observed_counts.query('observed >= min_counts')

If that cannot "see" `min_counts`, then perhaps try this instead:

    observed_freqs = observed_counts.query('observed >= {}'.format(min_counts))

aaron-mcdaid-zalando: I don't like this `dropna` here. As far as I can see, there cannot be any NAs here, as they were already filtered out from the data earlier.

aaron-mcdaid-zalando: (Actually, now I have a lot of questions. Is `observed_freqs` intended to be a DataFrame or a Series. Let's chat in person, @daryadedik . Just ping me any time!)

daryadedik: aha, with the changes above, we truly don't need it)

daryadedik: never used .query, looks nice)

aaron-mcdaid-zalando: I have my code for rerunning existing analyses through expan locally. Maybe this will help when you are ready to start making changes to ExpanService and wish to test them

Anyway, maybe this is needed. Note the `['observed']` on the end:

    observed_freqs = observed_counts.query('observed >= min_counts')['observed']
    assert isinstance(observed_freqs, pd.Series)

daryadedik: observed_freqs and expected_freqs are intended to be DataFrames, I made it for simplicity of working with data frame column and a dictionary of weights. I will try to think of better solution.

daryadedik: @aaron-mcdaid-zalando I removed that DataFrame conversion, now they are both Series. Should be a bit simpler now.

daryadedik:
@aaron-mcdaid-zalando let me know if you are ok with the changes.

expan/core/experiment.py
```diff
@@ -287,11 +287,56 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
+        observed_freqs = variants_counts[variants_counts >= min_counts]
+
+        # If there are less than 2 categories left after dropping counts less than 5 we can't conduct the test.
+        if len(observed_freqs) < 2:
+            raise ValueError("If the number of categories is less than 2 Chi-square test is not applicable.")
+
+        # Calculate expected counts given corresponding weights,
+        # weights are filtered out of categories which were dropped before.
+        total_count = observed_freqs.sum().sum()
```
aaron-mcdaid-zalando: if `observed_freqs` is a Series, then I think a single `.sum()` is sufficient

expan/core/experiment.py
```diff
@@ -287,11 +287,56 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        weights = {k: v for (k, v) in weights.items() if k in observed_freqs.index.values}
+        expected_freqs = pd.Series(weights)
+        expected_freqs *= total_count
+
+        # Compute chi-square and p-value statistics
+        chi_square_val, p_val = statx.chi_square(observed_freqs.sort_index(), expected_freqs.sort_index())
+
+        return p_val >= alpha, p_val, chi_square_val
```
aaron-mcdaid-zalando, line 342: I would like us to log the `variant_column.value_counts` also, so perhaps it could be returned here

Although perhaps this is something for ExpanService to log instead. In that case, I have no problem with this as it is now

expan/core/experiment.py
daryadedik: definitely!

aaron-mcdaid-zalando:
:+1:

PR #233, opened 2018-07-27
check if there are no p values to correct

gbordyugov:
This tries to handle the lack of p-value in a slightly more elegant way.

### DIFF
--- a/expan/core/correction.py
+++ b/expan/core/correction.py
@@ -12,8 +12,7 @@ def benjamini_hochberg(false_discovery_rate, original_p_values):
     :return: new critical value (i.e. the corrected alpha)
     :rtype: float
     """
-    if not original_p_values:
-        raise ValueError("empty array of p values to analyse")
+    assert original_p_values, "empty array of p values to analyse"
 
     p_values_sorted = np.sort(np.asarray(original_p_values))
     number_tests = len(original_p_values)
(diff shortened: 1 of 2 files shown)


expan/core/correction.py
```diff
@@ -3,15 +3,18 @@
...
     :param false_discovery_rate: proportion of significant results that are actually false positives
     :type  false_discovery_rate: float
     :param original_p_values: p values from all the tests
     :type  original_p_values: list[float]
-    
+
     :return: new critical value (i.e. the corrected alpha)
     :rtype: float
     """
+    if not original_p_values:
```
shansfolder: A few challenges from me ;)

In which cases will we pass an empty list of p values here?
Did you discover this problem from a real experiment bug?
Do we want to raise an exception or just log the error?

Thank you!

gbordyugov: @shansfolder that's something that you explained to me: if there are statistical tests with too few samples, they're marked as invalid and the overall number of statistical tests is decreased. So if all statistical tests in a statistical test suite are invalid, we wind up with having zero statistical tests to analyse (and zero original p values).

I haven't thought know about the logging yet.

igusher: @shansfolder  this is a great question!
We need to add input parameter validation in the call stack up to entry methods in expan lib (Experiment class I guess)

igusher: @gbordyugov I believe `assert` instruction is a bit more self-explanatory to check methods' pre-, post- conditions.
But I don't really mind, feel free to use it if you agree.

shansfolder: I see. Thank you. 
That's a log to me (otherwise we should raise the exception earlier when we find that there is no valid tests in the suite).

gbordyugov: @shansfolder for me, that's not just a log - it's unrecoverable. If there are no p-values to correct, we cannot return a sensible corrected alpha threshold.

shansfolder (approved):

PR #234, opened 2018-07-27
Re-wrote chi-square, removed dropping buckets.

daryadedik:
As discussed I removed category dropping (now we don't drop any categories), min_counts is input parameter (I set it to 5 as default), but in ADS afterward we will decide which threshold to use.
Also added number of filtered entities per variant in exp.metadata.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -337,7 +337,7 @@ class Experiment(object):
         expected_freqs = pd.Series(weights)
         expected_freqs *= total_count
 
-        #     chi-square and p-value statistics
+        # Compute chi-square and p-value statistics
         chi_square_val, p_val = statx.chi_square(observed_freqs.sort_index(), expected_freqs.sort_index())
 
         return p_val >= alpha, p_val, chi_square_val
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -319,14 +319,13 @@ class Experiment(object):
 
         # Count number of observations per each variant
         variant_column = pd.Series(variant_column).dropna(axis=0)
-        variants_counts = variant_column.value_counts()
+        observed_freqs = variant_column.value_counts()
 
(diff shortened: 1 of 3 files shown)


shansfolder:
@daryadedik I think I need a meeting in person to understand this PR ;)

expan/core/experiment.py
```diff
@@ -319,11 +319,12 @@ def chi_square_test_result_and_statistics(self, variant_column, weights, min_cou
...
-        variants_counts = variant_column.value_counts()
+        observed_freqs = variant_column.value_counts()
 
-        # Ensure at least a frequency of 5 at every location in observed_counts, otherwise drop the category
-        # see http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
-        observed_freqs = variants_counts[variants_counts >= min_counts]
+        # Ensure at least a frequency of 5 at every location in observed_counts.
+        # It's recommended to not conduct test if frequencies in each category is less than 5
+        if len(observed_freqs[observed_freqs < min_counts]) >= 1:
+            raise ValueError("Chi-square test is not valid for small expected or observed frequencies.")
```
igusher: @daryadedik @gbordyugov So what is a canonical way in Python to test input parameters?
asserts? or if-raise?

igusher: as I see it for a second time already, I tag you both

daryadedik:
Ok, as discussed I just returned previous chi-square implementation (nothing changed here), plus some small code beatifying made, and removed category dropping (now we don't drop any categories), min_counts is input parameter (I set it to 5 as default), but in ADS afterward we will decide which threshold to use.

daryadedik:
Sorry for mixing up the PR. I also added number of filtered entities per variant in exp.metadata. Will need that information instead computing the filtered number on ADS side.

shansfolder (approved):

PR #236, opened 2018-08-03
If the number of *finite* samples is too small, then the data isn't valid for analysis

aaron-mcdaid-zalando:
This is to fix a bug which the number of *finite* samples (non-NaN, non-inf) is too small.

(we should add units tests for this too)

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -89,6 +89,9 @@ class Experiment(object):
             data_for_analysis = feature.apply_to_data(data_for_analysis)
 
         if not self._is_valid_for_analysis(data_for_analysis, test):
+            # Note that this does not check that there are enough
+            # non-NaN and non=Inf datapoints. See below for a check
+            # of that:
             logger.warning("Data is not valid for the analysis!")
             return test_result
         if data_for_analysis.entity.duplicated().any():
@@ -104,6 +107,11 @@ class Experiment(object):
         logger.info("Treatment group size: {}".format(treatment.shape[0]))
         treatment_denominators = self._get_denominators(data_for_analysis, test, test.variants.treatment_name)
         treatment_numerators   = treatment * treatment_denominators
+
+        number_of_finite_controls   = np.sum(np.isfinite( control_numerators   / control_denominators   ))
+        number_of_finite_treatments = np.sum(np.isfinite( treatment_numerators / treatment_denominators ))
+
+        if number_of_finite_controls < 2 or number_of_finite_treatments < 2: return test_result
 
         # run the test method
         test_statistics = worker(x=treatment_numerators, y=control_numerators, x_denominators = treatment_denominators, y_denominators = control_denominators)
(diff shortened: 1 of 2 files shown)


aaron-mcdaid-zalando:
Fixing one bug somehow exposed other bugs :-)

In order to be able to count the number of `isfinite` entries, I had to coerce the data to `np.array`s of `float64`. On live, I think they are already like that, but the unit tests have some `np.arrays` of `object`; so I have to change them.

aaron-mcdaid-zalando:
... on second thoughts, seems like I can just adjust the test data slightly

shansfolder (approved):

aaron-mcdaid-zalando:
I guess we should add a test for this, and also perhaps move this logic into `_is_valid_for_analysis`?

aaron-mcdaid-zalando:
@gbordyugov , this is related to the ZeroDivisionError ticket

expan/core/experiment.py
```diff
@@ -105,6 +108,11 @@ def analyze_statistical_test(self, test, test_method='fixed_horizon', include_da
         treatment_denominators = self._get_denominators(data_for_analysis, test, test.variants.treatment_name)
         treatment_numerators   = treatment * treatment_denominators
 
+        number_of_finite_controls   = np.sum(np.isfinite( control_numerators   / control_denominators   ))
+        number_of_finite_treatments = np.sum(np.isfinite( treatment_numerators / treatment_denominators ))
+
+        if number_of_finite_controls < 2 or number_of_finite_treatments < 2: return test_result
```
gbordyugov, line 114: Is there maybe a better alternative than returning an empty result?

expan/core/experiment.py
```diff
@@ -265,10 +273,10 @@ def _is_valid_for_analysis(self, data, test):
 
     def _get_denominators(self, data, test, variant_name):
         if type(test.kpi) is not DerivedKPI:
-            return 1.0
+            return np.float64(1.0)
```
gbordyugov, line 276: why is this important?

expan/core/experiment.py
aaron-mcdaid-zalando: That seems to be the convention in this function; I'm just mirroring the behaviour of `_is_valid_for_analysis` (see a few lines above). But yes, I must I don't really know what the conventions are in Expan about what to return and when

expan/core/experiment.py
aaron-mcdaid-zalando: It might not be important actually!

In our internal software, we get errors like the following:

    TypeError: can't multiply sequence by non-int of type 'float'

This suggest a situation where there is confusion about whether to interpret `[3,4,5] * 2.0` as `[6,8,10]` or as `[3,4,5]*2 == [3,4,5,3,4,5]`. To avoid ambiguity, we should consider everything as a `numpy` type

aaron-mcdaid-zalando:
... to clarify my last comment, I should clarify that the `TypeError` happens in a different context and is not resolved by this PR. But that error did make me think that we need to be clearer about our types in our softwar

There are other approaches to the type system, I don't like this solution exactly, but this is one possible solution

expan/core/experiment.py
aaron-mcdaid-zalando: It might be better to move this logic into `_is_valid_for_analysis`

aaron-mcdaid-zalando:
We should add more unit tests for this issue. But we'll release now

PR #238, opened 2018-08-16
Stop warning about NaNs

aaron-mcdaid-zalando:
Stop warning about NaNs. They are perfectly legitimate datapoints, when using derived KPIs

Longer term, perhaps there is an appropriate place to print the numbers of NaNs in the *input* data, in a nicely-tabulated piece of text. But the current warning is very spammy

### DIFF
--- a/expan/core/statistics.py
+++ b/expan/core/statistics.py
@@ -113,15 +113,6 @@ def delta(x, y, x_denominators=1, y_denominators=1, assume_normal=True, alpha=0.
     del x_denominators
     del y_denominators
 
-    x_nan = np.isnan(_x_ratio).sum()
-    y_nan = np.isnan(_y_ratio).sum()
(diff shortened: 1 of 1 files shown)


shansfolder (approved):

gbordyugov:
@aaron-mcdaid-zalando could you pls elaborate on that? What happens if all of the data points are NaNs? Do we accept NaNs as valid values for means/std/p-values?

aaron-mcdaid-zalando:
@gbordyugov,

Currently, for means/std/p-values/sample-sizes, the NaNs are discarded. So the NaN values aren't really an issue (except that `expan` currently includes the NaNs when attempting to decide if there is a sufficient number of datapoints).

So, we should forget about the NaNs, and instead count the number of non-NaNs.

Longer term, a lot of this needs to be redone entirely though :-)

PR #239, opened 2018-09-04
Added observed and expected frequencies to chi-square statistics

daryadedik:
This is needed to log observed and expected frequencies.

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -318,7 +318,7 @@ class Experiment(object):
         :param alpha: significance level, 0.05 by default
         :return: True(if split is consistent with the given split) or
                  False(if split is not consistent with the given split)
-        :rtype:  Boolean, float, float
+        :rtype:  Boolean, float, float, pd.Series, pd.Series
         """
         if not hasattr(variant_column, '__len__'):
             raise ValueError("Variant split check was cancelled since input variant column is empty or doesn't exist.")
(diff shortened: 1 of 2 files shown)


expan/core/experiment.py
```diff
@@ -351,4 +351,4 @@ def chi_square_test_result_and_statistics(self, variant_column, weights, min_cou
         # Compute chi-square and p-value statistics
         chi_square_val, p_val = statx.chi_square(observed_freqs.sort_index(), expected_freqs.sort_index())
 
-        return p_val >= alpha, p_val, chi_square_val
+        return p_val >= alpha, p_val, chi_square_val, observed_freqs, expected_freqs
```
shansfolder, line 354: Is this going to be used in ADS?

shansfolder (approved):

expan/core/experiment.py
igusher: I'm still periodically getting surprised why you remind about ADS here.
Expan shouldn't care, it is a lib.

daryadedik: yes, in ADS we want to log the percentage difference between observed and expected counts 
along with the chi-square results to have more information about the split for the experiment. I don't want to duplicate the counts computation in ADS since we already have them here.

shansfolder: @igusher I know, I am asking just for my understanding. And Gute Besserung btw! ;)

PR #240, opened 2018-09-06
Changed chi-square test, removed frequencies computation

daryadedik:
In this PR I made some changes to previous chi-square helper method, which calculated observed frequencies from variant columns (which is very ADS related) and expected frequencies from weights. 
Now it accepts observed and expected frequencies as series and runs chi-square test if observed and expected frequencies are valid for the test (filter variants with the small number of observations and do not run test if the number of variants is less than 2).

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -306,49 +306,40 @@ class Experiment(object):
             flags = flags | data[column].apply(method_table[threshold_type])
         return flags
 
+    def run_goodness_of_fit_test(self, observed_freqs, expected_freqs, alpha=0.01, min_counts=5):
+        """ Checks the validity of observed and expected counts and runs chi-square test for goodness of fit.
 
-    def chi_square_test_result_and_statistics(self, variant_column, weights, min_counts=5, alpha=0.05):
-        """ Tests the consistency of variant split with the hypothesized distribution.
-        
-        :param variant_column: variant column from the input data frame
-        :param weights: dict with variant names as keys, weights as values
-                        ({<variant_name>:<weight>, ...}
-        :param min_counts: minimum number of observed and expected frequencies (should be at least 5), see 
-                            http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.chisquare.html
-        :param alpha: significance level, 0.05 by default
-        :return: True(if split is consistent with the given split) or
-                 False(if split is not consistent with the given split)
-        :rtype:  Boolean, float, float, pd.Series, pd.Series
+        :param observed_freqs: observed frequencies
+        :type  observed_freqs: pd.Series
+        :param expected_freqs: expected frequencies
+        :type  expected_freqs: pd.Series
+        :param alpha: significance level
+        :type  alpha: float
+        :param min_counts: minimum number of observations to run chi-square test
+        :type  min_counts: int
+        :return split_is_correct: False is split is biased and True if split is correct
+        :rtype: bool
         """
-        if not hasattr(variant_column, '__len__'):
-            raise ValueError("Variant split check was cancelled since input variant column is empty or doesn't exist.")
-        if not hasattr(weights, '__len__'):
-            raise ValueError("Variant split check was cancelled since input weights are empty or doesn't exist.")
-        if len(weights) <= 1 or len(variant_column) <= 1:
-            raise ValueError("Variant split check was cancelled since input weights or the number if categories "
-                             "is less than 2.")
 
-        # Count number of observations per each variant
-        variant_column = pd.Series(variant_column).dropna(axis=0)
-        observed_freqs = variant_column.value_counts()
+        if not hasattr(observed_freqs, '__len__'):
+            raise ValueError("Variant split check was cancelled since input observed frequencies are empty "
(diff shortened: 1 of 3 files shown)


daryadedik:
I left pd.Series for expected and observed freqs, because they are easy to work with, but if you think dict would be better, I can consider this.

expan/core/experiment.py
```diff
@@ -306,49 +306,37 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
-
-        # Compute chi-square and p-value statistics
-        chi_square_val, p_val = statx.chi_square(observed_freqs.sort_index(), expected_freqs.sort_index())
-
-        return p_val >= alpha, p_val, chi_square_val, observed_freqs, expected_freqs
+        valid_observed_freqs = observed_freqs[observed_freqs > min_counts]
+        valid_expected_freqs = expected_freqs.filter(valid_observed_freqs.keys())
+
+        if len(valid_observed_freqs) == len(valid_expected_freqs) and len(valid_observed_freqs) >= 2:
+            _, p_value = statx.chi_square(valid_observed_freqs.sort_index(), valid_expected_freqs.sort_index())
```
shansfolder, line 336: Very minor thing: ```statistics.py``` is imported with name ```statx``` and ```scipy.stats``` is imported with name ```stats``` (line 638 in ```statistics.py```). Maybe we can use the full name ```scipy.stats``` for less confusion.

expan/core/experiment.py
```diff
@@ -306,49 +306,37 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+            split_is_unbiased = p_value >= alpha
+        else:
+            raise ValueError("Variant split check was cancelled since input observed or expected frequencies "
```
shansfolder: \+ **valid** observed or expected frequencies?

expan/core/experiment.py
```diff
@@ -306,49 +306,37 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
-            raise ValueError("Variant split check was cancelled since input weights are empty or doesn't exist.")
-        if len(weights) <= 1 or len(variant_column) <= 1:
-            raise ValueError("Variant split check was cancelled since input weights or the number if categories "
-                             "is less than 2.")
 
-        # Count number of observations per each variant
-        variant_column = pd.Series(variant_column).dropna(axis=0)
-        observed_freqs = variant_column.value_counts()
+        if not hasattr(observed_freqs, '__len__'):
+            raise ValueError("Variant split check was cancelled since input observed frequencies are empty "
```
shansfolder: hmm... why not get the length and compare to 0? if it doesn't have ```'__len__'```, the input type is wrong I guess...

tests/tests_core/test_experiment.py
```diff
@@ -360,108 +360,86 @@ def test_outlier_filtering_derived_kpi(self):
...
-    def test_chi_square_test_result_and_statistics_counts_less_5_2_categories(self):
+    def test_run_goodness_of_fit_test_NaN_data(self):
         exp = self.getExperiment()
-        data = ['A'] * 17 + ['B'] * 2 + ['C'] * 3
-        weights = {'A': 0.33, 'B': 0.33, 'C': 0.33}
+        observed_freqs = pd.Series([170, 170], ['A', np.nan])
+        expected_freqs = pd.Series([170, 170], ['A', 'B'])
         with self.assertRaises(ValueError):
-            exp.chi_square_test_result_and_statistics(data, weights)
+            exp.run_goodness_of_fit_test(observed_freqs, expected_freqs)
```
shansfolder: 1. from which line will it raise a ValueError? Maybe you can also assert the error message so that we are sure it raises at the expected line.

2. Will it raise an error when:
```
observed_freqs = pd.Series([170, 170], ['A', 'C'])
expected_freqs = pd.Series([170, 170], ['A', 'B'])
```

expan/core/experiment.py
daryadedik: 👍

tests/tests_core/test_experiment.py
daryadedik: 1. yes, will add that
2. yes, it will, because I always filter expected_freqs by observed_freqs keys.

shansfolder: 2. cool. could you also add a test to confirm it?

expan/core/experiment.py
daryadedik: yes, I will check for input type and if the input is empty

expan/core/statistics.py
```diff
@@ -635,6 +635,6 @@ def chi_square(observed_freqs, expected_freqs, ddof=0):
     :return: chi-square statistics and p-value
     :rtype:  float, float
     """
-    chi_square_val, p_val = stats.chisquare(f_obs=observed_freqs, f_exp=expected_freqs, ddof=ddof, axis=None)
+    chi_square_val, p_val = scipy.stats.chisquare(f_obs=observed_freqs, f_exp=expected_freqs, ddof=ddof, axis=None)
```
daryadedik, line 638: here it's now clear it's from scipy.

shansfolder (approved):

PR #241, opened 2018-10-15
Make data for StatisticalTest optionally lazy

gbordyugov:
Make data for StatisticalTest optionally lazy

### DIFF
--- a/expan/core/statistical_test.py
+++ b/expan/core/statistical_test.py
@@ -19,6 +19,10 @@ class StatisticalTest(JsonSerializable):
     :type  variants: Variants
     """
     def __init__(self, data, kpi, features, variants):
+        # sometimes we want to postpone the actual creation of the
+        # dataframe
+        if callable(data):
+            data = data()
         if not isinstance(data, pd.DataFrame):
             raise ValueError("Please, provide data for the statistical test in form of a data frame.")
         if not isinstance(features, list):


expan/core/statistical_test.py
```diff
@@ -19,6 +19,10 @@ class StatisticalTest(JsonSerializable):
     :type  variants: Variants
     """
     def __init__(self, data, kpi, features, variants):
+        # sometimes we want to postpone the actual creation of the
+        # dataframe
+        if callable(data):
+            data = data()
```
shansfolder, line 25: I am still not sure if you get my point. 

Given a feature with multiples values, say "Browser" which have "Chrome", "FF" and "IE". This will create 3 `StatisticalTest` objects where you pass the same `callable` function to it. 
```
"StatisticalTest(bla, data, browser=chrome)"
"StatisticalTest(bla, data, browser=ff)"
"StatisticalTest(bla, data, browser=ie)"
```

From our experience, we know this `data()` function to create dataframe is runtime-expensive (groupby, multi-variance-drop, outlier-filtering, etc). We want to call it only once for the feature "browser" (instead of repeating 3 times for "chrome", "ie" and "ff"), because it creates the same dataframe for browser.

e.g. if `data()` is already called on `a = StatisticalTest(bla, data, browser=chrome)`, we don't want to call it again on `b =StatisticalTest(bla, data, browser=ie)`, but somehow want to use the already computed one from object `a`.

This will become a major bottleneck of runtime, because with `date` and `appdomain` as subgroup, we will have lots of repeated calls to the expensive `data()` function.

how can you ensure that? Is it the smart counter example we discussed? 
Maybe I didn't understand you, but to me it's not trivial.

shansfolder: The current implementation in `ExpanService` computes the dataframe only once per subgroup. One for `browser`, one for `date`, one for `appdomain` and so on.

gbordyugov: @shansfolder thanks for the detailed rebuttal.

This is only the Expan side of the problem. On the client side (ExpanService), we could make `data`, for example, caching:

```
def dataframe_maker(args):
  df = [None] # we need a list here because of how closures in Python work
  def make_df():
    if not df[0]:
      df[0] = load_dataframe_do_group_by_etc(args)
    return df[0]
  return make_df
```

And then use it in different `StatisticalTest`s as

```
# first make the maker with the arguments, just one time
df_maker = dataframe_maker(browser_sga_args)

# use it in as many places as one wishes:

# in one place
df1 = df_maker()

# in some other place
df2 = df_maker()

# in a third place
df3 = df_maker()
```

The logic above would ensure that `load_dataframe_do_group_by_etc(args)` is called only once despite `df_maker()` being called three times.

What do you think of it?

PR #242, opened 2018-11-26
Fix early stopping, by making it stricter

aaron-mcdaid-zalando:
The implementation of "alpha spending" increased our Type I error rate from 5% to approximately 10%, as it didn't spend alpha 'piece-wise' but instead it spent it 'cumulatively'

Therefore, this PR makes the early stopping stricter by an arbitrary factor of 100. As a result, we will get an 'extra Type I error rate' of approximately 0.05% instead of 5%.

Then, the total Type I error rate will be 5.05% (= 5%+0.05%) instead of  10% (=5%+5%)

### DIFF
--- a/expan/core/early_stopping.py
+++ b/expan/core/early_stopping.py
@@ -33,6 +33,8 @@ def obrien_fleming(information_fraction, alpha=0.05):
     :return: redistributed alpha value at the time point with the given information fraction
     :rtype:  float
     """
+
+    alpha = alpha/100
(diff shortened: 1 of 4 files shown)


shansfolder (approved):

aaron-mcdaid-zalando:
In practice though, we are unlikely to use this internally without further investigation of the implications. In the last few minutes we had a meeting about this option, and other options, for proceeding

Therefore, we won't immediately merge this

daryadedik:
👍

daryadedik:
👍

daryadedik:
👍

PR #245, opened 2019-01-07
Correct documentation

shansfolder:
From what I saw in the codebase, we have only implemented [equal variance t-test](https://en.wikipedia.org/wiki/Student%27s_t-test#Equal_or_unequal_sample_sizes,_equal_variance), not welch's t-test.

We first compute the pooled std [here]( https://github.com/zalando/expan/blob/master/expan/core/statistics.py#L308), and then multiply the pooled std with the square root of ` (1/n1 + 1/n2)` [here](https://github.com/zalando/expan/blob/master/expan/core/statistics.py#L482), which is exactly the formula of equal variance t-test in the link above. As far as I see, I couldn't find implementation of welch's t-test in ExpAn.

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -12,9 +12,10 @@ Assumptions used in analysis
   * Variance in treatment and control is identical
   * Mean of delta is normally distributed
 
-2. Welch t-test
+2. Equal or unequal sample sizes, equal variance t-test
 
   * Mean of means is t-distributed (or normally distributed)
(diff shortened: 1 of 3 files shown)


shansfolder:
Ready for review. 

There is also a package version conflict because of the [fresh release 13h ago](https://pypi.org/project/pytest-cov/#history) (Jan 7 2019!) of `pytest-cov 2.6.1.

gbordyugov (approved):

daryadedik (approved):

PR #246, opened 2019-01-22
refactor get_data_for_analysis

igusher:
My main goal is to change responsibilities a bit between classes in order to modularize computations so that later we can reuse some chunks in distributed environment. 

The first thing I want to do - is to encapsulate computational logic in StatisticalTest. As it will be a building block in distributed environment and one StatisticalTest can run independently from the other. (Of course later StatisticalTestSuite will collect all Tests and correct them)
....
And I will do it in baby-steps so that you can see if I break any existing assumptions.

PR #247, opened 2019-01-23
Refactoring/move analysis logic inside statistical test

igusher:
Now the meat of it.

PR is based on another one - so there are duplicated changed.
https://github.com/zalando/expan/pull/246


igusher:
Temporarily closing it, to avoid confusion, let's review #246 first to reduce efforts by doing it in small chunks.

PR #249, opened 2019-06-28
Two-sided outlier filtering mode

gbordyugov:
Two-sided outlier filtering mode

### DIFF
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -377,8 +377,9 @@ class Experiment(object):
 
 def _choose_threshold_type(data):
     """ Heuristics used to decide what filtering method to use."""
-    assert data.min, "data should have min method"
-    assert data.max, "data should have max method"
+    assert data,     'data should be non-empty'
+    assert data.min, 'data should have min method'
+    assert data.max, 'data should have max method'
     min, max = data.min(), data.max()
 
     if min < 0.0 and max > 0.0:
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -301,14 +301,17 @@ class Experiment(object):
         """
 
         def find_smallest(data, quantile):
+            """ Return indices of data points smaller than given quantile."""
             threshold = data.quantile(quantile)
             return data.apply(lambda x: x < threshold)
 
         def find_largest(data, quantile):
+            """ Return indices of data points larger than given quantile."""
             threshold = data.quantile(quantile)
             return data.apply(lambda x: x > threshold)
 
         def find_smallest_and_largest(data, quantile):
+            """ Return indices of data points outside of the given quantile."""
             rest = 1.0 - quantile
             quantiles = [rest/2.0, 1.0 - rest/2.0]
             thresholds = list(data.quantile(quantiles))
@@ -365,3 +368,14 @@ class Experiment(object):
             raise ValueError("Variant split check was cancelled since observed or expected frequencies "
                              "are less than 2.")
         return split_is_unbiased, p_value
+
+def _choose_threshold_type(data):
+    """ Heuristics used to decide what filtering method to use."""
+    min, max = data.min(), data.max()
+
+    if min < 0.0 and max > 0.0:
+        return 'two-sided'
+    elif max > 0.0:
+        return 'upper'
+    elif min < 0.0:
+        return 'lower'
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -309,28 +309,27 @@ class Experiment(object):
             return data.apply(lambda x: x > threshold)
 
         def find_smallest_and_largest(data, quantile):
-            print(data)
             rest = 1.0 - quantile
             quantiles = [rest/2.0, 1.0 - rest/2.0]
-            print(quantiles)
             thresholds = list(data.quantile(quantiles))
-            print(thresholds)
             return data.apply(lambda x: x < thresholds[0] or x > thresholds[1])
 
         quantile = float(percentile)/100.0
+
+        method_table = {'upper': find_largest,
+                        'lower': find_smallest,
+                        'two-sided': find_smallest_and_largest}
 
         for col in data[kpis].columns:
             column = data[col].copy()
             column.replace([np.inf, -np.inf], np.nan)
 
-            if threshold_type == 'upper':
-                flags = flags | find_largest(column, quantile)
-            elif threshold_type == 'lower':
-                flags = flags | find_smallest(column, quantile)
-            elif threshold_type == 'two-sided':
-                flags = flags | find_smallest_and_largest(column, quantile)
-            else:
+            if threshold_type not in method_table:
                 raise ValueError("Unknown outlier filtering method '%s'."%(threshold_type,))
+            else:
+                method = method_table[threshold_type]
+                flags = flags | method(column, quantile)
+
         return flags
 
     def run_goodness_of_fit_test(self, observed_freqs, expected_freqs, alpha=0.01, min_counts=5):
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -302,17 +302,17 @@ class Experiment(object):
         """
 
         def find_smallest(data, quantile):
-            """ Return indices of data points smaller than given quantile."""
+            """ Return boolean vector of data points smaller than given quantile."""
             threshold = data.quantile(quantile)
             return data.apply(lambda x: x < threshold)
 
         def find_largest(data, quantile):
-            """ Return indices of data points larger than given quantile."""
+            """ Return boolean vector of data points larger than given quantile."""
             threshold = data.quantile(quantile)
             return data.apply(lambda x: x > threshold)
 
         def find_smallest_and_largest(data, quantile):
-            """ Return indices of data points outside of the given quantile."""
+            """ Return boolean vector of data points outside of the given quantile."""
             rest = 1.0 - quantile
             quantiles = [rest/2.0, 1.0 - rest/2.0]
             thresholds = list(data.quantile(quantiles))
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -13,7 +13,7 @@ from expan.core.results import StatisticalTestResult, MultipleTestSuiteResult, C
 warnings.filterwarnings("ignore", category=FutureWarning)
 logger = logging.getLogger(__name__)
 
-
+DEFAULT_OUTLIER_QUANTILE = 0.99
 class Experiment(object):
     """ Class which adds the analysis functions to experimental data. """
     def __init__(self, metadata):
@@ -330,7 +330,7 @@ class Experiment(object):
                 threshold_type, quantile = thresholds[col]
                 quantile = quantile/100.0
             else:
-                quantile = 0.99
+                quantile = DEFAULT_OUTLIER_QUANTILE
                 threshold_type = _choose_threshold_type(column)
 
             if threshold_type not in method_table:
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -198,10 +198,8 @@ class Experiment(object):
         
         :param kpis: list of KPI instances
         :type  kpis: list[KPI]
-        :param percentile: percentile considered as filtering threshold
-        :type  percentile: float
         :param thresholds: dict of thresholds mapping KPI names to (type, percentile) tuples
-        :type  threshold_type: dict
+        :type  thresholds: dict
 
         :return: Will return data with filtered outliers.
         """
@@ -296,10 +294,8 @@ class Experiment(object):
         
         :param kpis: the kpis to perform filtering
         :type  kpis: list[str]
-        :param percentile: percentile considered as filtering threshold
-        :type  percentile: float
-        :param threshold_type: type of threshold used ('upper', 'lower', or 'two-sided')
-        :type  threshold_type: str
+        :param thresholds: dict of thresholds mapping KPI names to (type, percentile) tuples
+        :type  thresholds: dict
 
         :return: boolean values indicating whether the row should be filtered
         :rtype: pd.Series
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -222,9 +222,9 @@ class Experiment(object):
         if 0.0 < percentile <= 100.0 is False:
             raise ValueError("Percentile value needs to be between 0.0 and 100.0!")
         # check if provided filtering kind is valid
-        admittable_thresholds = set(['upper', 'lower', 'both'])
+        admittable_thresholds = set(['upper', 'lower', 'two-sided'])
         if threshold_type not in admittable_thresholds:
-            raise ValueError("Threshold type needs to be either 'upper', 'lower', or 'both'.")
+            raise ValueError("Threshold type needs to be either 'upper', 'lower', or 'two-sided'.")
 
         # run quantile filtering
         flags = self._quantile_filtering(data=data,
@@ -310,31 +310,28 @@ class Experiment(object):
             return data.apply(lambda x: x > threshold)
 
         def find_smallest_and_largest(data, quantile):
+            print(data)
             rest = 1.0 - quantile
             quantiles = [rest/2.0, 1.0 - rest/2.0]
+            print(quantiles)
             thresholds = list(data.quantile(quantiles))
+            print(thresholds)
             return data.apply(lambda x: x < thresholds[0] or x > thresholds[1])
 
         quantile = float(percentile)/100.0
 
         for col in data[kpis].columns:
-            column = data[col]
-            min, max = column.min(), column.max()
-
-            # if max > 0.0 and min < 0.0:
-            #     flags = flags | find_smallest_and_largest(column, quantile)
-            # elif max > 0.0:
-            #     flags = flags | find_largest(column, quantile)
-            # elif min < 0.0:
-            #     flags = flags | find_smallest(column, quantile)
+            column = data[col].copy()
+            column.replace([np.inf, -np.inf], np.nan)
 
             if threshold_type == 'upper':
                 flags = flags | find_largest(column, quantile)
             elif threshold_type == 'lower':
                 flags = flags | find_smallest(column, quantile)
-            else: # threshold_type == 'two_sided':
+            elif threshold_type == 'two-sided':
                 flags = flags | find_smallest_and_largest(column, quantile)
-            # flags = flags | find_largest(column, quantile)
+            else:
+                raise ValueError("Unknown outlier filtering method '%s'."%(threshold_type,))
         return flags
 
     def run_goodness_of_fit_test(self, observed_freqs, expected_freqs, alpha=0.01, min_counts=5):
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -2,6 +2,7 @@ import logging
 import warnings
 
 import numpy as np
+import pandas as pd
 import copy
 
 import expan.core.early_stopping as es
@@ -14,6 +15,7 @@ warnings.filterwarnings("ignore", category=FutureWarning)
 logger = logging.getLogger(__name__)
 
 DEFAULT_OUTLIER_QUANTILE = 0.99
+
 class Experiment(object):
     """ Class which adds the analysis functions to experimental data. """
     def __init__(self, metadata):
@@ -391,9 +393,8 @@ class Experiment(object):
 def _choose_threshold_type(data):
     """ Heuristics used to decide what filtering method to use."""
     assert len(data), 'data should be non-empty'
-    assert data.min,  'data should have min method'
-    assert data.max,  'data should have max method'
-    min, max = data.min(), data.max()
+    data = pd.Series(data)
+    min, max = data.min(skipna=True), data.max(skipna=True)
 
     if min < 0.0 and max > 0.0:
         return 'two-sided'
--- a/expan/core/experiment.py
+++ b/expan/core/experiment.py
@@ -327,7 +327,7 @@ class Experiment(object):
                 of data."""
             rest = 1.0 - quantile
 
-            neg_threshold = data[data <= 0.0].quantile(rest/2.0)
+            neg_threshold = data[data <  0.0].quantile(rest/2.0)
             pos_threshold = data[data >= 0.0].quantile(1.0 - rest/2.0)
 
             return data.apply(lambda x: x < neg_threshold or x > pos_threshold)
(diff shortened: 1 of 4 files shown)


igusher:
LGTM.
Approve.

shansfolder (approved):

gbordyugov:
@aaron-mcdaid-zalando what do you think?

expan/core/experiment.py
```diff
@@ -339,3 +374,14 @@ def run_goodness_of_fit_test(self, observed_freqs, expected_freqs, alpha=0.01, m
             raise ValueError("Variant split check was cancelled since observed or expected frequencies "
                              "are less than 2.")
         return split_is_unbiased, p_value
+
+def _choose_threshold_type(data):
```
igusher, line 393: preconditions, mostly for documentation? 
```assert data, "data should be defined"```

Comment (gbordyugov) [reply expan/core/experiment.py:393]: good point, thank you, see below.

File: expan/core/experiment.py [hunk 2]
```diff
@@ -290,20 +294,51 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        method_table = {'upper': find_largest,
+                        'lower': find_smallest,
+                        'two-sided': find_smallest_and_largest}
+
+        for col in data[kpis].columns:
+            column = data[col].copy()
+            column.replace([np.inf, -np.inf], np.nan)
+
+            quantile = 0.99
+            threshold_type = _choose_threshold_type(column)
```
igusher: reassignment below is a bit confusing, maybe this can go to an `else` branch. 
Though, super minor

expan/core/experiment.py
igusher: `data = np.array([])` will throw on `data.min()`

gbordyugov: ... aka developers playing compiler.

aaron-mcdaid-zalando:
Can you say more about what this is about. I recall us having a discussion internally some time ago, but I would like it specified somewhere. Perhaps as documentation.

The default is the compute the X percentile (where X is 99.9 by default for some users) and to discard values above that threshold

The change proposed here is:
 1. Compute percentile X for all _non-negative_ values and discard above that.
 2. Compute percentile ~X~ (1-X) for all _negative_ values, and to discard values below

This scheme is proposed in order to have backwards compatibility for datasets that do not have any negative values - correct?

gbordyugov:
@aaron-mcdaid-zalando three methods that are implemented do the following:
 1) drop x% of the largest values
 2) drop x% of the smallest values
 3) drop x/2% on both sides of the distribution.

Plus there is this heuristics that goes for 1) if all values are non-negative, for 2) if all values are non-positive, and for 3) if the values are sowohl als auch.

expan/core/experiment.py
```diff
@@ -190,18 +190,16 @@ def analyze_statistical_test_suite(self, test_suite, test_method='fixed_horizon'
         return test_suite_result
 
 
-    def outlier_filter(self, data, kpis, percentile=99.0, threshold_type='upper'):
+    def outlier_filter(self, data, kpis, thresholds=None):
         """ Method that filters out entities whose KPIs exceed the value at a given percentile.
         If any of the KPIs exceeds its threshold the entity is filtered out.
```
aaron-mcdaid-zalando, line 197: This changes the interface and therefore will break existing code for some users (i.e. us!).

Do we care about this? @shansfolder @daryadedik 

We could make changes to ensure compatibility

gbordyugov: The other parts of the highly classified Z. code should be changed accordingly, but fortunately, it's just one call.

expan/core/experiment.py
```diff
@@ -218,26 +216,32 @@ def outlier_filter(self, data, kpis, percentile=99.0, threshold_type='upper'):
...
 
-        # check if provided percentile is valid
-        if 0.0 < percentile <= 100.0 is False:
-            raise ValueError("Percentile value needs to be between 0.0 and 100.0!")
-        # check if provided filtering kind is valid
-        if threshold_type not in ['upper', 'lower']:
-            raise ValueError("Threshold type needs to be either 'upper' or 'lower'!")
+        admittable_thresholds = set(['upper', 'lower', 'two-sided'])
+
+        thresholds = thresholds or {}
```
aaron-mcdaid-zalando, line 224: Why `thresholds = thresholds or {}`?

`thresholds` is expected to be a `dict` anyway. I don't see what this line does

expan/core/experiment.py
```diff
@@ -218,26 +216,32 @@ def outlier_filter(self, data, kpis, percentile=99.0, threshold_type='upper'):
...
+        for kpi in thresholds:
```
aaron-mcdaid-zalando: Maybe `for kpi in thresholds.keys():` ?

aaron-mcdaid-zalando: or even:

```
    for kpi_name, (threshold_type, percentile) in thresholds.items():
```

expan/core/experiment.py
gbordyugov: `thresholds` has `None` as a default value. `a = a or []/{}/set()` is a common idiom to resolve `None` to the default empty container.

aaron-mcdaid-zalando: I was about to say: _"Shouldn't the default be `{}` then?"_

but then I remembered that python default values can have weird lifetimes!

I'll resolve!

expan/core/experiment.py
aaron-mcdaid-zalando: I'll 'resolve' this. It's true that we don't need to ensure compatibility here

expan/core/experiment.py
```diff
@@ -290,20 +294,51 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
         :rtype: pd.Series
         """
-        method_table   = {'upper': lambda x: x > threshold, 'lower': lambda x: x <= threshold}
-        na_replacement = {'upper': -float_info.max,         'lower':  float_info.max}
 
-        for column in data[kpis].columns:
-            threshold = np.percentile(data[column].fillna(na_replacement[threshold_type]), percentile)
-            flags = flags | data[column].apply(method_table[threshold_type])
+        def find_smallest(data, quantile):
+            """ Return indices of data points smaller than given quantile."""
```
aaron-mcdaid-zalando: Instead of "indices", should we say a "boolean vector"?

I'm asking because it took me quite some time to understand the existing code (and I found (and fixed) a bug) while doing so.

gbordyugov: done

expan/core/experiment.py
```diff
@@ -290,20 +294,51 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+            """ Return indices of data points outside of the given quantile."""
+            rest = 1.0 - quantile
+            quantiles = [rest/2.0, 1.0 - rest/2.0]
+            thresholds = list(data.quantile(quantiles))
+            return data.apply(lambda x: x < thresholds[0] or x > thresholds[1])
+
+        method_table = {'upper': find_largest,
+                        'lower': find_smallest,
+                        'two-sided': find_smallest_and_largest}
+
```
aaron-mcdaid-zalando, line 339: (Just a question). So `two-sided` is the symmetric case

aaron-mcdaid-zalando: ... oh. Now I realise that you are building support into expan for both the symmetric case and the asymmetric case!

gbordyugov: 1. tá.

2. Not quite, it's either of three. Not possible right now to apply `upper` on the positive ones and `lower` on the negative ones without writing additional code on the client's side.

expan/core/experiment.py
```diff
@@ -290,20 +294,51 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+            column = data[col].copy()
+            column.replace([np.inf, -np.inf], np.nan)
+
+            if col in thresholds:
+                threshold_type, quantile = thresholds[col]
+                quantile = quantile/100.0
+            else:
+                quantile = 0.99
+                threshold_type = _choose_threshold_type(column)
+
```
aaron-mcdaid-zalando, line 350: This `else` is interesting. It's a default threshold?

Perhaps there should be a `DEFAULT_OUTLIER_QUANTILE=0.99` at the top of the file?

gbordyugov: done

tests/tests_core/test_experiment.py
```diff
@@ -288,37 +288,41 @@ def test_analyze_statistical_test_with_nan_data(self):
...
         self.assertEqual(filtered_dict['B'], 298)
 
+    def test_outlier_filtering_automatic(self):
+        exp = self.getExperiment()
+        data = exp.outlier_filter(
+            self.data,
+            kpis = [KPI(kpi) for kpi in self.kpi_names],
+        )
+        self.assertEqual(len(self.data) - len(data), exp.metadata['filtered_entities_number'])
+        self.assertEqual(exp.metadata['filtered_entities_number'], 389)
```
gbordyugov, line 318: Note that this value is slightly larger than 386 from line 305 due to some values being negative and `two-sided` test being applied.

expan/core/experiment.py
```diff
@@ -339,3 +374,19 @@ def run_goodness_of_fit_test(self, observed_freqs, expected_freqs, alpha=0.01, m
...
+    min, max = data.min(), data.max()
+
+    if min < 0.0 and max > 0.0:
+        return 'two-sided'
+    elif max > 0.0:
+        return 'upper'
+    elif min < 0.0:
+        return 'lower'
+    else:
+        raise ValueError("Both min and max of data are zero.")
```
aaron-mcdaid-zalando: We shouldn't raise an error here. We could quite legitimately have vectors full of zeroes, especially for small datasets

In general, we should treat `0` (and `np.inf`, `-np.inf` and `np.nan`) as legitimate values. Our derived KPIs will often be `0/0 = np.nan`

How do we expect outlier filtering to interact with `np.nan`? If we have 1'000'000 values, of which 300'000 are `np.nan`, and we have 99% as our threshold, will we remove 10'000 or 3'000?

gbordyugov: I'm returning 'two-sided' here now, too.

Comparison of a nan and any normal number always returns `False`, so they're never taken out.

tests/tests_core/test_experiment.py
aaron-mcdaid-zalando, line 318: This test says it removes `389`, but the one about finds `386` to remove. Is there a reason for this difference?

It seems to be the same data, the same KPIs, and the same (default) threshold of 99%. Is it a difference in the threshold_type?

aaron-mcdaid-zalando: Doh! I see my question was already answered

Resolving

tests/tests_core/test_outliers.py
```diff
@@ -1,14 +1,75 @@
...
+def test_choose_threshold_type_lower_zero():
+    data = np.array([-3, -2, -1, 0])
+    assert _choose_threshold_type(data) == 'lower'
+
+
+def test_choose_threshold_type_two_sided():
+    data = np.array([-3, -2, -1, 0, 1, 2, 3])
+    assert _choose_threshold_type(data) == 'two-sided'
+
+
```
aaron-mcdaid-zalando, line 31: Let's add a test with a `nan`, e.g. `data = np.array([-3, np.nan, -1])`

(This comment is related to my comment on the implementation of `_choose_threshold_type`)

I made such a test on my laptop and I see the message: `ValueError: Both min and max of data are zero.`. That message isn't true of course, but it seems to be how `np.nan` are interpreted in the current implementation

gbordyugov: I don't throw this error anymore

tests/tests_core/test_outliers.py
```diff
@@ -1,14 +1,75 @@
...
+    assert flags.tolist() == [False]*2 + [True]*5 + [False]*5
+
+
+def test_quantile_filtering_two_sided():
+    exp = Experiment({})
+    df = pd.DataFrame.from_dict({'earnings' : list(range(10))})
+
+    flags = exp._quantile_filtering(df, ['earnings'], {'earnings': ('two-sided', 80.0)})
+    results = flags.tolist()
+    assert results == [True] + [False]*8 + [True]
```
aaron-mcdaid-zalando, line 75: In Google Chat, you said _"the asymmetric case is not yet implemented"_

This test tells me otherwise

aaron-mcdaid-zalando: I forgot to add a smiley there !!

Have I misunderstood something?

gbordyugov: That's the symmetric case.

By saying symmetric, I mean that 0.5% of the whole distribution are cropped off on the right and the left side.

Asymmetric means that 0.5% of non-positive values are cropped off on the left and 0.5% of non-negative are cropped off on the right.

expan/core/experiment.py
```diff
@@ -218,26 +216,33 @@ def outlier_filter(self, data, kpis, percentile=99.0, threshold_type='upper'):
...
-        # check if provided filtering kind is valid
-        if threshold_type not in ['upper', 'lower']:
-            raise ValueError("Threshold type needs to be either 'upper' or 'lower'!")
+        admittable_thresholds = set(['upper', 'lower', 'two-sided',
+                                     'two-sided-asym'])
+
+        thresholds = thresholds or {}
+        for kpi in thresholds.keys():
+            threshold_type, percentile = thresholds[kpi]
+
```
aaron-mcdaid-zalando, line 227: `for kpi, (threshold_type, percentile) in thresholds.items():` is another possibility

(I thought I made this comment already, but I can't find it. Sorry for any repetition, and sorry if I missed any response)

expan/core/experiment.py
```diff
@@ -290,20 +295,63 @@ def _quantile_filtering(self, data, kpis, percentile, threshold_type):
...
+        def find_smallest_and_largest_asym(data, quantile):
+            """ Return boolen vector of data to remove such that quantile/2
+                is kept in both non-negative and non-positive subsets
+                of data."""
+            rest = 1.0 - quantile
+
+            neg_threshold = data[data <= 0.0].quantile(rest/2.0)
+            pos_threshold = data[data >= 0.0].quantile(1.0 - rest/2.0)
+
```
aaron-mcdaid-zalando, line 332: `neg_threshodl` should be based on `<` instead of `<=`

And maybe `pos` should be renamed `non_negative`

expan/core/experiment.py
gbordyugov: You have suggested `for kpi in thresholds.keys():` before, indeed.

expan/core/experiment.py
gbordyugov: I'd like to keep `<=` and not to rename the variables :~) There is indeed a slight disagreement between the names and the semantics, but I think that introducing smth like `non_negative` will result in an overhead.

expan/core/experiment.py
aaron-mcdaid-zalando: > tá

[Go raibh maith agat (thanks!)](https://en.wiktionary.org/wiki/go_raibh_maith_agat)

This list of options makes more sense now that I see the `two-sided-asym` option - it completes the set

expan/core/experiment.py
```diff
@@ -339,3 +387,20 @@ def run_goodness_of_fit_test(self, observed_freqs, expected_freqs, alpha=0.01, m
...
+    min, max = data.min(), data.max()
+
+    if min < 0.0 and max > 0.0:
+        return 'two-sided'
+    elif max > 0.0:
+        return 'upper'
+    elif min < 0.0:
+        return 'lower'
+    else:
+        # it doesn't really matter since in this case min == max == 0.0
```
aaron-mcdaid-zalando, line 406: If there are any `NaN`s in the data, it will reach this `else`. So I think the comment is incorrect

It's a bit awkward here though. I would suggest `data.min(skipna=True)` as this works for `pandas.Series`, but it doesn't work for `numpy` arrays. So I don't know what types and methods we can rely on here

Ideally, we would do one of these things in this `_choose_threshold_type` function:
 - remove the nan
 - or replace the nans with zero

But, as this is just the default behaviour for columns without an explicitly-requested threshold_type, maybe we don't need to worry about this

tests/tests_core/test_outliers.py
aaron-mcdaid-zalando: There should be at least one test with a `nan` in it. Whatever `NaN`-behaviour we have should be tested (and documented?)

Currently, it would simply be something like: 

```
def test_choose_threshold_type_with_nan():
    # confirm that any data with any 'np.nan' in it will result in 'two-sided'
    data = np.array([-3, -2, -1, np.nan])
    assert _choose_threshold_type(data) == 'two-sided'
```

tests/tests_core/test_outliers.py
aaron-mcdaid-zalando: Resolving. I see now that I had mixed up symmetry and assymetry

expan/core/experiment.py
gbordyugov: done, sticking to `Series`.

tests/tests_core/test_outliers.py
gbordyugov: Added `nan`s to tests.

expan/core/experiment.py
aaron-mcdaid-zalando: Doh! I thought of both, but maybe I only mentioned one of them. 

Anyway, I'll resolve this now

expan/core/experiment.py
aaron-mcdaid-zalando: In a dataset with zeroes, `<=` will result in too many values being removed

While there might be a lot of debate about which values to remove, we should be consistent on how many values are removed. I.e. That percentile _p_ will result in _(1-p)*n_ values being removed. This requires that we have non-overlapping predicates here

~(Although, I admit that I still don't know what to do with NaNs)~

expan/core/experiment.py
aaron-mcdaid-zalando: Nice. Resolving.

expan/core/experiment.py
aaron-mcdaid-zalando: Aha. It seems like we're moving towards the following rule, based on the `skipna` thing in `_choose_threshold_type`:

"Given `n` datapoints - of which `f` are finite (i.e. neither `nan` nor `(-)inf`) - and a quantile `q`, we will remove `f*(1-q)` of them (rounded down the the nearest integer). The threshold_type will control which ones to remove, but won't affect how many are removed"

Is that correct?

tests/tests_core/test_outliers.py
aaron-mcdaid-zalando: Nice. Resolving

expan/core/experiment.py
gbordyugov: For upper/lower/two-sides -- yes.

For two-sides-asym it applies to both non-negative and non-positive ones separately with q' = q/2.

aaron-mcdaid-zalando: So in the extreme case where everything is zero (we have some KPIs which have lots of zeroes in them), we will remove `2*f*(1-q)` of them

gbordyugov: Is what you're writing a desired behaviour? Because

```
In [1]: from pandas import Series
In [2]: s = Series([0.0]*10)
In [3]: s.quantile([i * 0.1 for i in range(10)])
Out[3]:
0.0    0.0
0.1    0.0
0.2    0.0
0.3    0.0
0.4    0.0
0.5    0.0
0.6    0.0
0.7    0.0
0.8    0.0
0.9    0.0
dtype: float64
```

aaron-mcdaid-zalando: > ... desired behaviour?

One issue is that I don't really know what I think the desired behaviour is. Maybe a Speclog would help here

In fact, I think I'll make such a document right now, and ask for feedback on it.

aaron-mcdaid-zalando: ... although, on second thoughts, I won't take that initiative right now. Let's ask around

gbordyugov: In the updated code line, line 330 reads
```
neg_threshold = data[data <  0.0].quantile(rest/2.0)
```

aaron-mcdaid-zalando (approved):
I like the fact this this is being extended to include multiple different approaches to outlier filtering. This allows users (including ourselves) to choose their own preferences later

gbordyugov:
@aaron-mcdaid-zalando could you pls review the changes?

expan/core/experiment.py
aaron-mcdaid-zalando: Thanks for changing that. Now, with `data<0` and `data>=0`, we have non-overlapping sets used for defining the thresholds

tests/tests_core/test_outliers.py
```diff
@@ -5,27 +5,27 @@
...
 def test_choose_threshold_type_lower_zero():
-    data = np.array([-3, -2, -1, 0])
+    data = np.array([-3, -2, -1, 0, np.nan])
     assert _choose_threshold_type(data) == 'lower'
 
 
 def test_choose_threshold_type_two_sided():
-    data = np.array([-3, -2, -1, 0, 1, 2, 3])
+    data = np.array([-3, -2, -1, 0, 1, 2, 3, np.nan])
     assert _choose_threshold_type(data) == 'two-sided'
```
aaron-mcdaid-zalando, line 29: Good to see `NaN`s in the tests, and `skipna` to ensure they are ignored when deciding the threshold_type

expan/core/experiment.py
gbordyugov: can I resolve this?

expan/core/experiment.py
gbordyugov: can I resolve it?

aaron-mcdaid-zalando (approved):
(adding another 'Approve' review, after recent changes)

The thresholds are now computed on separate, non-overlapping, subsets of the data (negative and non-negative)

As stated in my previous review, this PR enables four different methods for selecting outliers for removal. This allows the users a little more flexibility in deciding what policy to use. Great feature, implemented well!

expan/core/experiment.py
aaron-mcdaid-zalando: I'll resolve. The conversion to `pandas.Series` helps

aaron-mcdaid-zalando:
> @aaron-mcdaid-zalando could you pls review the changes?

Conversations resolved, and another positive review added. :+1:

PR #250, opened 2019-09-16
Added 4eye principle for approving PRs

daryadedik:
Also added maintainers and contributors. 
According to https://opensource.zalando.com/docs/releasing/index/#be-compliant

### DIFF
--- a/.zappr.yaml
+++ b/.zappr.yaml
@@ -1,5 +1,5 @@
-X-Zalando-Type: tools
 X-Zalando-Team: octopus
+X-Zalando-Type: code
 
 approvals:
   minimum: 2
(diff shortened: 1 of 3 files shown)


.zappr.yaml
```diff
@@ -1 +1,17 @@
 X-Zalando-Type: tools
```
shansfolder: it should not be `tools` I guess

daryadedik: should be "code"? I noticed a lot of team either using "tools" or not including this type at all.

daryadedik: I think "code" should be ok.

daryadedik:
👍

shansfolder:
👍

daryadedik:
👍

daryadedik:
👍

PR #256, opened 2021-03-31
Update requirements.txt

andompesta:
pystan moved from version 2.* to 3.* 
As such the naming is changed and creates some import errors. Version 2.19.* seems to work fine


shansfolder:
thanks @andompesta for the contribution! The test failed after the new pystan version because we hard coded the numerical test values before, which is not right way to test. I will take a look in a few days.

